PR #143, opened 2015-03-25
Apropos

floren:
This lets you do shorthand commands, similar to Cisco's command line.

You only have to type enough to match a pattern. For example, although minimega has the commands "vm cdrom" and "vm config", it is sufficient to type "vm c" to run "vm config", because "vm cdrom" requires additional arguments and thus cannot match "vm c". Similarly, "vm c e 0" matches "vm cdrom eject 0", because "vm config" does not register any patterns which would match "vm c e 0".


floren:
Didn't realize you can only assign one person...

djfritz:
LGTM

PR #148, opened 2015-03-31
Ticket 86

jcrussell:
Getting rid of ugly vmConfigFns. Replacing with tags/reflection. Added some unit tests to make sure it's sane but could use more testing.


djfritz:
@floren can you comment - I think I'm out of the loop here. I think I see the rationale, but please correct me if I'm missing a larger issue with the VM code (of which there are many... sorry):

Most of the 'vm config' fields are simple strings or integers, with default values equal to the 'zero' value for that type. That opens up the possibility of applying some form of generics on the 'vm config' fields by figuring out if they're string/int/bool, etc. and controlling them that way. There are cases where the 'vm config' fields are more than simple type, and as such need their own special cases. In the current implementation, /every/ 'vm config' field gets a set/clear/print function. In the new implementation, only the 'special' fields do. 

Is that right? If not, then perhaps disregard the rest because I'm sure I'm just being a boofhead here.

These changes work, but I'm having a little trouble seeing why the old configFns were a bad thing. 

It looks like the change nearly doubled the LOC, imports 'reflect' (which Rob and the Go community say to avoid unless strictly necessary), and adds custom cases in several places (but much fewer total custom cases, since essentially every configFn was a custom case). The old configFns centralized code for a given 'vm config' option. That meant that if I wanted to add a new config, I would add a field to the vmInfo struct, add the configFns for it, and add the minicli handler wrapper. 

Now I have to potentially add corner case checks, depending on the config field, in 4 places across 3 files in addition to the minicli handler and vmInfo struct changes. 

The ticket calls for simplifying the VM code around this - it seems more complicated now. 

At any rate, let's only merge 2.0 tickets and related bits until 2.0 is released.

jcrussell:
> Most of the 'vm config' fields are simple strings or integers, with default values equal to the 'zero' value for that type. That opens up the possibility of applying some form of generics on the 'vm config' fields by figuring out if they're string/int/bool, etc. and controlling them that way. There are cases where the 'vm config' fields are more than simple type, and as such need their own special cases. In the current implementation, /every/ 'vm config' field gets a set/clear/print function. In the new implementation, only the 'special' fields do.

> Is that right? If not, then perhaps disregard the rest because I'm sure I'm just being a boofhead here.

That's right, it automatically handles the common case for string/bool/[]string types.

> It looks like the change nearly doubled the LOC, imports 'reflect' (which Rob and the Go community say to avoid unless strictly necessary), and adds custom cases in several places (but much fewer total custom cases, since essentially every configFn was a custom case). The old configFns centralized code for a given 'vm config' option. That meant that if I wanted to add a new config, I would add a field to the vmInfo struct, add the configFns for it, and add the minicli handler wrapper.

Good points, some thoughts:

 * It's actually fewer LOC if you don't include the 188 LOC for new unit tests
 * I thought quite a bit about the use of reflect here. The alternative was lots of boilerplate cases in switch statements. I think using reflect for the tags feature is probably the most accepted use of reflection.
 * It's a little more complicated to add custom cases but adding the common cases is three easy steps: add a handler, add a field with tag to the vmInfo struct, and add it's name to vmConfigFields.

floren:
What do we want to do on this one?

djfritz:
Hold on with this. It's not a 2.0 priority and doesn't change any
functionality.

PR #150, opened 2015-03-31
Fix #149.

### LINKED ISSUE #149: minimega: panic on 'read' with whitespace
root@ccc1:~# cat > foo.mm << EOF
> echo foo
> 
> echo bar
> EOF
@jcrussell @floren 

root@ccc1:~# ./minimega

minimega$ read foo.mm 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x43eb24]

goroutine 28 [running]:
main.hasCommand(0x0, 0x949f70, 0x4, 0x0)
        /home/fritz/repos/minimega/src/minimega/misc.go:292 +0xf4
main.cliRead(0xc20804f950, 0xc2080645a0)
        /home/fritz/repos/minimega/src/minimega/misc_cli.go:142 +0x887
minicli.func·009()
        /home/fritz/repos/minimega/src/minicli/minicli.go:123 +0x69
created by minicli.ProcessCommand
        /home/fritz/repos/minimega/src/minicli/minicli.go:132 +0x214

jcrussell:
Need to check if cmd is nil before running hasCommand.

### DIFF
--- a/src/minimega/misc_cli.go
+++ b/src/minimega/misc_cli.go
@@ -138,16 +138,16 @@ func cliRead(c *minicli.Command, respChan chan minicli.Responses) {
 			break
 		}
 
-		// HAX: Make sure we don't have a recursive read command
-		if hasCommand(cmd, "read") {
-			err = errors.New("cannot run nested `read` commands")
-			break
-		}
-
 		// No command was returned, must have been a blank line or a comment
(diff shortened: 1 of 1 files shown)


djfritz:
LGTM

PR #152, opened 2015-04-01
Finish package building process

floren:
This is the last bit to complete the script for building a Debian package. Just run the script from the packaging/debian directory and it will create a .deb


floren:
@djfritz

djfritz:
LGTM - We should also figure out how we want to distribute this stuff at some point.

Issue #153, opened 2015-04-01
minicli: suport != in .filter builtin


djfritz:
@jcrussell 

See summary. We should update the API doc on this as well.

Last issue before 2.0 feature freeze!

djfritz:
closed by https://github.com/sandia-minimega/minimega/pull/155

PR #155, opened 2015-04-02
minicli: updated .filter, new API.

jcrussell:
Added inverted filter (field!=value). Also added a `MustCompile` method
that runs CompileCommand and calls log.Fatal if there was an error.

Fixes issue #153.


djfritz:
LGTM

Issue #144, opened 2015-03-25
minimega: default gui/web paths


floren:
"web" needed the path to novnc. We defaulted to $pwd/misc/novnc.

"gui" needs novnc, d3, and presumably eventually a javascript terminal. It defaults those to /opt/minimega/misc/[novnc,d3]

I've generally held that $pwd/misc/novnc is a decent start because less experienced users will probably be running "./bin/minimega" from the root of the repo, while more advanced/dedicated users who have created /opt/minimega can be expected to change paths themselves if they need.

Jeremy suggested an "env" command, which we could just use to set $CWD, or we could set a "webroot" variable which is used as the root for novnc, d3, and other web junk we need.

bjwrigh:
I can go for the "env" command.  I plan to overtake the 'web' code soon, as
well as co-locate novnc and web things (if nobody opposes either) along
with some of my code review and overhaul as I roll it into master.

floren:
The GUI looks really good, so I'd like to bring it into master ASAP.

Here's what I propose as a simple way to bring in "gui":

* Move misc/novnc and misc/d3 into misc/web
* Change "gui" command to take a webroot argument, so you say "gui webroot /opt/minimega/misc/web"
* Make default webroot $CWD/misc/web
* Remove old "web" code, rename "gui" to "web".

Advantages:

* Less configuration needed
* Default webroot helps newbies who are probably running from inside the minimega dir

djfritz:
Let's put up a pull request for this.

For the environment - the default http.Transport registers paths with Handler() and HandlerFunc(), but doesn't have a facility to change those paths once you start the webserver. That stinks if you accidentally put the wrong path in, or use the default and the files aren't there. You either have to restart minimega or move/link files.

Can we add a custom http.Transport or whatever other interface is needed to make the paths mutable at runtime?

djfritz:
Is this done in the new web pull request?

djfritz:
fixed by https://github.com/sandia-minimega/minimega/issues/156

Issue #140, opened 2015-03-25
New GUI


djfritz:
See summary - Much work has been done on a new web interface, as shown in the gui branch. Finish, polish, and merge to master.

floren:
TESTING
?       apigen  [no test files]
?       gonetflow       [no test files]
?       gopcap  [no test files]
--- FAIL: TestDecodeRawRGB (0.00s)
        reader_test.go:89: open testdata/test_rgb_raw.ppm: no such file or directory
FAIL
FAIL    gopnm   0.002s

djfritz:
failed tests are fixed with:

https://github.com/sandia-minimega/minimega/commit/eb62ea83f2052a612608965c4e8c670e3142a366

Please pull from master.

djfritz:
fixed by https://github.com/sandia-minimega/minimega/issues/156

PR #156, opened 2015-04-02
Brave New Web

jcrussell:
This is the new web code that was started by Brian. I'm hoping it makes the 2.0 release as it's pretty nice and it changes the web CLI.  I've refactored the code quite a bit and removed some of the parts that aren't relevant to general minimega users (we need to have a discussion on the VM runtime status). Also, I removed the web shell.

The only tests I've done are on a single machine with a single VM. I plan to test with a few machines and a dozen VMs today.

Feedback would be appreciated. What's missing? Can something be improved?


djfritz:
Files changes = 139... good thing I took my soma this morning...

djfritz:
I think with a pull this large and 2.0 looming, and no known deliverables that don't already require the new gui, we just do a cursory review and test of the request, merge it if there are no glaring issues, and test for bugs as they come in over the next week of testing master.

I'll pull this and test on ccc this morning.

src/minimega/web.go
```diff
@@ -1,50 +1,76 @@
...
+	Server    *http.Server
+	Templates *template.Template
+	Port      int
+}
 
 var webCLIHandlers = []minicli.Handler{
 	{ // web
-		HelpShort: "start the minimega web interface",
+		HelpShort: "start the minimega webserver",
 		HelpLong: `
```
djfritz: We need to say what the default path and port are in the help text.

djfritz: Woop, I see the default path IS in there, we still need the port.

djfritz:
If I start the web service without a working web template directory under it I get an error. That's fine, but it's not very descriptive. The error instead should be about how it cannot find the web directory. 

Also, if I try to browse to the host after starting the web service without the web directory in place, even if I put the web directory in place, I get a panic.

minimega$ web /does/not/exist
minimega$ 2015/04/02 17:44:05 ERROR web.go:114: web: couldn't initalize templates: html/template: pattern matches no files: `misc/web/templates/*.html`

minimega$ 2015/04/02 17:44:40 http: panic serving [::1]:36730: runtime error: invalid memory address or nil pointer dereference
goroutine 4747 [running]:
net/http.func·011()
        /opt/go/src/net/http/server.go:1130 +0xbb
html/template.(*Template).lookupAndEscapeTemplate(0x0, 0x998790, 0x9, 0x0, 0x0, 0x0)
        /opt/go/src/html/template/template.go:99 +0x72
html/template.(*Template).ExecuteTemplate(0x0, 0x7f543b1b9460, 0xc208094140, 0x998790, 0x9, 0x0, 0x0, 0x0, 0x0)
        /opt/go/src/html/template/template.go:88 +0x54
main.webRenderTemplate(0x7f543b1b9428, 0xc208094140, 0x998790, 0x9, 0x0, 0x0)
        /home/djfritz/repos/real_minimega/src/minimega/web.go:157 +0xa4
main.webHome(0x7f543b1b9428, 0xc208094140, 0xc2081ac000)
        /home/djfritz/repos/real_minimega/src/minimega/web.go:236 +0x59
net/http.HandlerFunc.ServeHTTP(0xa7d8e0, 0x7f543b1b9428, 0xc208094140, 0xc2081ac000)
        /opt/go/src/net/http/server.go:1265 +0x41
net/http.(*ServeMux).ServeHTTP(0xc20827e2d0, 0x7f543b1b9428, 0xc208094140, 0xc2081ac000)
        /opt/go/src/net/http/server.go:1541 +0x17d
net/http.serverHandler.ServeHTTP(0xc20851c300, 0x7f543b1b9428, 0xc208094140, 0xc2081ac000)
        /opt/go/src/net/http/server.go:1703 +0x19a
net/http.(*conn).serve(0xc208058500)
        /opt/go/src/net/http/server.go:1204 +0xb57
created by net/http.(*Server).Serve
        /opt/go/src/net/http/server.go:1751 +0x35e

src/minimega/vm.go
```diff
@@ -512,3 +512,50 @@ func ParseVmState(s string) (VmState, error) {
 
 	return VM_ERROR, fmt.Errorf("invalid state: %v", s)
 }
+
+// Get the VM info from all hosts optionally applying column/row filters.
+// Returns a map with keys for the hostnames and values as the tabular data
+// from the host.
+func globalVmInfo(masks []string, filters []string) map[string][][]string {
```
djfritz: We're no longer doing this right? Instead we're just letting vm info pack the vms.vms (or whatever it needs to be) into the Data element of the minicli response?

src/minimega/vmtags.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (2015) Sandia Corporation.
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+
+// This is based on https://golang.org/src/text/template/parse/lex.go
+
+package main
```
djfritz: can we replace this whole file with just packing []*vmInfo into the Data field of the response?

src/minimega/vmtags_test.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (2015) Sandia Corporation.
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+
+package main
```
djfritz: Same as previous comment.

djfritz:
Do we have a default view for the main page that isn't just the nav bar? Should we have a help text or something? A logo?

djfritz:
See above comments, let's do another review after those have been addressed. I've been trying to break it on the test cluster, and other than the bug noted above, it seems to work well.

djfritz:
Wondering if issue updates can be made with comments here:

updates #144 
updates #140

djfritz:
fixes #144

djfritz:
fixes #140

src/minimega/vmtags.go
jcrussell: Yup, we will.

src/minimega/vm.go
jcrussell: We may still want to keep this helper as it makes it pretty simple to pull info from all hosts. With vm info packing vms.vms into the Data element, this may change to return map[string]vms.

djfritz: This still seems web specific. If so, we should move it to web.go. I also sort of see this as a wrapper to runGlobally("vm info") or whatever the syntax for that is. If it's reduced to that, we should eliminate it.

jcrussell: dot will also use it.

djfritz:
confirmed panic noted above fixed by https://github.com/jcrussell/minimega/commit/ec522254b0fbc8e8d23da0fd5870f4f50623ab5a

djfritz:
LGTM, said with trepidation...

PR #163, opened 2015-04-02
describe python API changes, plus small tweak

devinc:
The python scripts no longer have to be run inside of the misc/python/ directory, and I updated the release document.


djfritz:
LGTM

PR #166, opened 2015-04-03
vmList struct -> VMs map

### LINKED ISSUE #164: minimega: remove vmList struct, promote vms to a typedef to the map
see summary. This is very old code that needs to be done away with. We have the list of vms in a map, which is then wrapped by a struct as the only member. Do away with the wrapper struct and promote the map to a typedef.

jcrussell:
Promoted vms from struct to typedef. Renamed vmList to VMs.

Fixes #164.


floren:
LGTM

PR #167, opened 2015-04-03
documentation: define naming conventions and apply them across the repo

floren:
We need to use the same capitalization and naming conventions everywhere. I created naming.article to define those conventions, then went through and enforced them everywhere.


floren:
@jcrussell @djfritz

djfritz:
LGTM

Issue #170, opened 2015-04-06
VM Tags Page


jcrussell:
When Brian created the gui branch, he had a page that displayed VMs with specific tags (traceroute, ...). We removed this from the merged code because it was too specific to his use case for general minimega. Therefore, I propose we make a generic "tags" page in web that allows the user to select and display VMs with the specified tags. The page would have a column on the left with a list of all the tags set across all VMs and checkboxes next to each tag name. When the user selects a set of the tags, she can hit "find" and a table of all the VMs with those tags (and their values) will be displayed in a table to the right of the tag selection column.

If implemented such that the tag names can be populated from GET variables, Brian could have a URL bookmarked that includes the four or so columns he's interested in.

Time estimate: 2 hours.

djfritz:
Sounds good to me. Do we want to make mutable on this page too, or do we wait to figure out a more general purpose interface into controlling an environment from the gui?

jcrussell:
For now, this will be read-only.

floren:
A patch has been merged, let's close this and get 1 step closer to 2.0

PR #172, opened 2015-04-07
minimega: added new vm tags page.

jcrussell:
Initial work on the vm tags page. The page displays all VMs and then you can
filter to VMs that have a set of tags selected on the left column. The
checkboxes can be set via URL parameters too.

Let me know what you think... Does this make sense? The equivalent URL for
Brian's old guiState would be: 

    http://localhost:9001/tags?traceroute&SNMP&DNS&app

Updates #170.


djfritz:
LGTM

PR #174, opened 2015-04-08
documentation: added a tutorial on vmbetter

floren:
A little tutorial that tells you how to use vmbetter to make a custom livecd. I did it as a livecd because that's useful outside the context of VMs, but in the end we still use minimega to test the ISO.


djfritz:
LGTM

PR #176, opened 2015-04-09
minimega: dirty hacks.

### LINKED ISSUE #175: Missing VM fields in new web
ip/ip6/cc_active are always blank in the new VM all table, these are generated by querying vm state from datastructures that are not present on the head node.

jcrussell:
Work arounds for not all VM state being stored in the VM struct.
Fallback on the Tabular data from minicli.Response to populate VM info
fields that are dynamically generated (ip, ip6, cc_active). This code
should be deleted as soon as the state for these is moved into the VM
struct.

Fixes #175


djfritz:
LGTM

Issue #187, opened 2015-04-28
minicli: .columns header mismatch


djfritz:
header mismatch error is happening specifically with .columns builtin where it shouldn't:

scenario: launch 1 vm on a node, 0 vms on another node. Run the following from a third node:

    mesh send all vm info

which works, and:

    .columns id,state mesh send all vm info

which throws the header mismatch error

@jcrussell

PR #192, opened 2015-04-29
minicli: fixes 187.

jcrussell:
.columns wasn't altering the Header for responses where Tabular was nil
which lead to mismatched Header columns. Changed it so that we always
update the Header (if it's not nil).


djfritz:
LGTM

djfritz:
fixes #187

PR #200, opened 2015-05-08
streaming responses to python

### LINKED ISSUE #197: minimega: python bindings send() not checking More boolean
Responses as defined in the localResponse struct, https://github.com/sandia-minimega/minimega/blob/master/src/minimega/cli.go#L52, Contain a More boolean that indicates to the client that it should continue reading for more responses. The expected behavior is that each localResponse should be rendered/printed upon receiving, and if the More field is set, read more data. This allows for things like the streaming responses that are printed when issuing a 'read' command. 

Currently only 'read' uses More, but the behavior should be future proofed by operating in send()

@devinc 

See https://github.com/sandia-minimega/minimega/blob/master/src/minimega/cli.go#L303 for the implementation of handling More with -attach and -e

devinc:
Fixes #197. Check the test case to see how it works.

@djfritz thoughts?


djfritz:
@bjwrigh 
@jcrussell 

Can you both take a look? The gist here is to support the More bool in the response to support things like 'read' and any future streaming responses.

jcrussell:
Looks like it should work unless the response has a "}" embedded in it.

devinc:
We discussed this offline already, but just for posterity: the '}' check is just an optimization so that it doesn't call json.loads() for every byte read from the socket (it's in a try/except). There may be a better way to do it.

djfritz:
So LGTM?

jcrussell:
Yup, LGTM.

Issue #191, opened 2015-04-28
Feature (that could fix a bug):  vm tag all {key} {value}


bjwrigh:
'mesh send all vm tag all time_started aug5_22:01'

add functionality of grouping / all to vm tag to allow the same tags to be added to multiple vms simultaneously.

bjwrigh:
the bug might've been attributed to kvmi things...assume no problem unless updated.

jcrussell:
This is pretty easy to implement. We could add the following patterns to vm tag:

vm tag <vm id or name or all> <key or all>
vm tag <vm id or name or all> <key> [value]

The first would allow you to get a table with rows: vm name, key, value. You could then filter this if needed.

The second does what the ticket describes.

@djfritz  -- [Y/n]?

djfritz:
Is it more consistent to keep the current "vm tag <name or id> <key> [value]" syntax and add a wildcard check to the name/id field?

jcrussell:
Updated comment.. angle brackets were filtered. Does that change your comment?

djfritz:
Ahh yes, you and I had the same idea. This looks good.

jcrussell:
Closing, fixed in #208.

PR #208, opened 2015-05-12
Lone vm ranger

### LINKED ISSUE #186: minimega: support ranges and lists for vm commands
'vm lock vm-0' would allow a vm to be immune to 'vm kill all'  in order to facilitate a special vm to maintain state while cycling the environment around the locked VMs.

jcrussell:
Support for list ranges in a few different commands. Much better than the recent movie by a similar name.

Closes #186, #191.

### DIFF
--- a/src/minimega/misc.go
+++ b/src/minimega/misc.go
@@ -6,18 +6,15 @@ package main
 
 import (
 	"crypto/rand"
-	"errors"
 	"fmt"
 	_ "gopnm"
 	"image"
 	"image/png"
 	"io/ioutil"
-	"math"
 	"minicli"
 	log "minilog"
 	"os"
 	"os/exec"
-	"ranges"
 	"regexp"
 	"resize"
 	"strconv"
@@ -311,66 +308,6 @@ func isReserved(s string) bool {
 	}
 
 	return false
-
-}
-
-// expandListRange takes a string such as "foo,bar[1-3]" and expands it to a
-// fully enumerated list of names.
-func expandListRange(in string) ([]string, error) {
-	var res, parts []string
-
-	var prev int
-	var inside bool
-
-	for i := 0; i < len(in); i++ {
-		if in[i] == '[' {
-			if inside {
-				return nil, fmt.Errorf("nested '[' at char %d", i)
-			} else {
-				inside = true
-			}
-		} else if in[i] == ']' {
-			if inside {
-				inside = false
-			} else {
-				return nil, fmt.Errorf("unmatched ']' at char %d", i)
-			}
-		} else if in[i] == ',' {
-			if !inside {
-				parts = append(parts, in[prev:i])
-				prev = i + 1
-			}
-		}
-	}
-
-	// handle last entry on the line and look for unterminated ranges
-	if inside {
-		return nil, errors.New("unterminated '['")
-	} else if prev < len(in)-1 {
-		parts = append(parts, in[prev:])
-	}
-
-	log.Debugln("expandListRange parts: ", parts)
-
-	for _, v := range parts {
-		index := strings.IndexRune(v, '[')
-		if index == -1 {
-			res = append(res, v)
-			continue
-		}
-
-		prefix := v[:index]
-		r, _ := ranges.NewRange(prefix, 0, math.MaxInt32)
-		ret, err := r.SplitRange(v)
-		if err != nil {
-			return nil, err
-		}
-		log.Debug("expanded range: %v", ret)
-		res = append(res, ret...)
-	}
-	log.Debugln("expandListRange res: ", res)
-
-	return res, nil
 }
 
 // hasWildcard tests whether the lookup table has Wildcard set. If it does, and
--- a/src/minimega/misc.go
+++ b/src/minimega/misc.go
@@ -6,15 +6,18 @@ package main
 
 import (
 	"crypto/rand"
+	"errors"
 	"fmt"
 	_ "gopnm"
 	"image"
 	"image/png"
 	"io/ioutil"
+	"math"
 	"minicli"
 	log "minilog"
 	"os"
 	"os/exec"
+	"ranges"
 	"regexp"
 	"resize"
 	"strconv"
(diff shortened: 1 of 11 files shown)


src/minimega/vm_cli.go
```diff
@@ -583,10 +584,41 @@ Clear all tags from all VMs:
 
         clear vm tag all`,
 		Patterns: []string{
-			"clear vm tag <vm id or name or all> [tag]",
+			"clear vm tag",
+			"clear vm tag <vm target> [tag]",
 		},
 		Call: wrapSimpleCLI(cliClearVmTag),
 	},
+	{ // vm target (dummy command)
```
djfritz: @floren 

This is an interesting approach, but does this add confusion to someone looking at the output of 'help'? Do they initially see 'vm target' as a command?

Do we accomplish the same thing by having the following one-liner in the help text of each command that has the 'target' syntax:

"vm target can be any of a VM name, ID, 'all', or a range such as foo,bar[1-3]"

src/minimega/misc.go
```diff
@@ -300,3 +303,81 @@ func isReserved(s string) bool {
 	return false
 
 }
+
```
djfritz: Does all of this fit better as a helper function in src/ranges?

djfritz:
Looks good overall. I'll start poking at it on one of the clusters.

djfritz:
BUG - started some vms, killed them, tried to start them again - it seems maybe they started on the first try but nothing got updated in the state?

    minimega$ .columns name,state vm info
    host | name  | state
    ccc1 | bar   | QUIT
    ccc1 | foo10 | RUNNING
    ccc1 | foo11 | QUIT
    minimega$ vm start bar
    Error (ccc1): VM bar is already running
    minimega$

djfritz:
similar behavior with 'vm start all':

    minimega$ vm launch foo[1-3]
    minimega$ .columns name,state vm info
    host | name | state
    ccc1 | foo1 | BUILDING
    ccc1 | foo2 | BUILDING
    ccc1 | foo3 | BUILDING
    minimega$ vm kill all
    minimega$ .columns name,state vm info
    host | name | state
    ccc1 | foo1 | QUIT
    ccc1 | foo2 | QUIT
    ccc1 | foo3 | QUIT
    minimega$ vm start all
    minimega$ .columns name,state vm info
    host | name | state
    ccc1 | foo1 | QUIT
    ccc1 | foo2 | QUIT
    ccc1 | foo3 | QUIT
    minimega$

djfritz:
vm tag seems to always throw an error, even if the operation is successful:

    minimega$ vm launch foo[1-5]
    minimega$ vm tag foo[1-2] foo bar
    Error (ccc1): VMs not found: [foo1 foo2]
    minimega$ .column name,tags vm info
    host | name | tags
    ccc1 | foo1 | map[foo:bar]
    ccc1 | foo2 | map[foo:bar]
    ccc1 | foo3 | map[]
    ccc1 | foo4 | map[]
    ccc1 | foo5 | map[]
    minimega$

jcrussell:
Thanks for all the bugs! I think we should have mm scripts and expected outputs checked into the repo somewhere, perhaps these will serve as a good starting point.

It shouldn't take too long to fix these tomorrow.

bjwrigh:
are you suggesting 'vm start all' will now start QUIT VMs?? @djfritz

jcrussell:
@bjwrigh is right, the reason the start commands weren't working is because you didn't pass the optional `quit` string.

djfritz:
There's a 'quit' string? Bah humbug. If I specify a specific VM to start - like the first bug comment above, it should start it. In fact, I think it is, because if I call start a second time on that VM, it nags that it's already started.

Also we shouldn't throw an error on calling vm start on a started vm - it should just return silently.

djfritz:
I think 'vm start all' should be the only special case - it should only start VMs in the paused or building state. If we specify a specific range or individual VM, it should also start quit/error VMs. That way we can do what @bjwrigh originally asked for in the original issue - a way to cycle lots of vms with just 2 commands:

    vm kill router[1-100]
    vm start router[1-100]

djfritz:
That also means ditching the 'quit' suffix.

jcrussell:
I definitely would *not* merge this... I want to do a lot more testing of the concurrency code.

src/minimega/misc.go
jcrussell: Are you referring to `expandListRange`? If so, yeah, we could move it.

djfritz:
Changes look sane - I'll grab some nodes and poke at it today.

djfritz:
The ID match/expansion seems odd to me:

    minimega$ vm launch foo[1-5]
    minimega$ .columns id,name,state vm info
    host | id | name | state
    ccc2 | 0  | foo1 | BUILDING
    ccc2 | 1  | foo2 | BUILDING
    ccc2 | 2  | foo3 | BUILDING
    ccc2 | 3  | foo4 | BUILDING
    ccc2 | 4  | foo5 | BUILDING
    minimega$ vm launch [0-4]
    minimega$ .columns id,name,state vm info
    host | id | name | state
    ccc2 | 0  | foo1 | BUILDING
    ccc2 | 1  | foo2 | BUILDING
    ccc2 | 2  | foo3 | BUILDING
    ccc2 | 3  | foo4 | BUILDING
    ccc2 | 4  | foo5 | BUILDING
    ccc2 | 5  | 0    | BUILDING
    ccc2 | 6  | 1    | BUILDING
    ccc2 | 7  | 2    | BUILDING
    ccc2 | 8  | 3    | BUILDING
    ccc2 | 9  | 4    | BUILDING
    minimega$ vm start [0-4]
    minimega$ .columns id,name,state vm info
    host | id | name | state
    ccc2 | 0  | foo1 | RUNNING
    ccc2 | 1  | foo2 | RUNNING
    ccc2 | 2  | foo3 | RUNNING
    ccc2 | 3  | foo4 | RUNNING
    ccc2 | 4  | foo5 | RUNNING
    ccc2 | 5  | 0    | BUILDING
    ccc2 | 6  | 1    | BUILDING
    ccc2 | 7  | 2    | BUILDING
    ccc2 | 8  | 3    | BUILDING
    ccc2 | 9  | 4    | BUILDING

We probably shouldn't allow launching numeric named VMs...

djfritz:
LGTM - did you still want to hold off on merging? If not, merge away.

PR #215, opened 2015-05-19
Integration Testing

jcrussell:
Beginnings of integration testing. Opening PR so that we can start discussing what things we should test.


djfritz:
This is neat. When we first started talking about adding runtime tests, we had a different scope of how it would be done - we would have yet another external tool that would launch minimega and do things. 

I wonder if an external tool/framework is still the better way to go for a couple of reasons:

* A framework test tool can test more than just minimega - vmbetter, etc.
* It can allow us to do assertions on things that aren't printed output from minimega, for example launching a vm with 'vm launch foo' may return nothing (expected), but we may want to check for a running kvm process with the args we expected. An external framework could be written to support asserting something like `ps aux | grep kvm...` == "kvm --args..."
* An external test tool could also support the output assertions provided here

There is at least one thing this could do that an external tool couldn't: 

* Internal assertions which could be provided with an 'integration' build, such as a "want" file that also asks to check for the correct internal state of some variable.

Thoughts?

jcrussell:
I'm not sure which way would be better. I see testing other tools like vmbetter and rfbplay as a much lower priority -- they are mostly feature complete and have been tested through repeated use. For minimega itself, it has been the focus of continuous development (at least for me) for almost eight months (wow, that went quick). Under the hood, there have been numerous changes that could have broken any number of use cases. We definitely need *something* to ensure it is behaving correctly as I continue to refactor and improve things. I'm especially nervous about bugs in the new VM Interface branch.

All that said, I think #215 provides a simple way to write test cases without too much additional work. We could generate a few dozen test cases in a day that provide a simple safety net. Later, if we decide that we need a standalone tool, we can always make use of the existing tests.

djfritz:
I agree. There may be some future where we want a bigger tool, but there's no reason it can't incorporate this (or in the worst case we just remove the integration build files and go about our business with the other tool). 

A few questions:

* What's the use case here for multi-node tests?
* How do we orchestrate launching tests with this? Another bash script?
* Does/can the orchestration tool restart minimega after each test (it probably should)?

jcrussell:
1. Can we leverage the deploy command for multi-node tests? We could also have a separate flag to say whether to run single vs multi node tests.
2. Right now, it's a one-liner but another bash script (or a flag on test.bash) probably makes sense.
3. I think it should, perhaps we can add a `restart` command to minimega? Can we just call teardown, and then fork/exec a new minimega? We may have to do a better job closing things (and use SO_REUSEPORT).

jcrussell:
Going to integrate the integration testing into the VM++ branch.

PR #221, opened 2015-05-28
Why so serial?

jcrussell:
Merging Jeremy's old serial code (after a few cleanups).

Diff'd the output from:
```
vm config virtio-serial 40
vm config qemu-override
```

Before and after my code cleanup -- no change. I plan to add a test to #215.

@djfritz

### DIFF
--- a/src/minimega/cc.go
+++ b/src/minimega/cc.go
@@ -123,7 +123,7 @@ func ccSerialWatcher() {
 
 	for {
 		// get a list of every vm's serial port path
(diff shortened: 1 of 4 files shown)


djfritz:
This looks pretty good to me. I'm going to test against the `cc serial` stuff, which expects at least one virtio port to be available. I think having 0 devices by default is the right thing to do, and have `cc serial` not freak out if a VM doesn't have a serial port enabled.

djfritz:
LGTM

PR #223, opened 2015-05-28
Fixes #222.

### LINKED ISSUE #222: minimega: keep internal api calls out of the cli history
Currently internal api calls like `vm info` and `vm screenshot` from the web interface get injected into the cli history. That clutters things when you go to do a `write` command after using the web interface.

The internal API calling infrastructure should stay out of the history.

jcrussell:
Added new built-ins: `.record` that controls whether subcommands get
recorded in the history. This allows us to `mesh send` commands and omit
them from the local history, as needed for things like web.


djfritz:
LGTM

PR #224, opened 2015-05-28
Bug fix.

jcrussell:
Overwrote cmd so cmd was nil inside error check (and thus we couldn't
access cmd.Original) which led to a panic. Fixed format string to be in
the right order.

### DIFF
--- a/src/minimega/cli.go
+++ b/src/minimega/cli.go
@@ -84,14 +84,16 @@ func runCommand(cmd *minicli.Command) chan minicli.Responses {
 // channel. This is useful if you want to get the output of a command from all
 // nodes in the cluster without having to run a command locally and over
 // meshage.
-func runCommandGlobally(cmd *minicli.Command, record bool) chan minicli.Responses {
+func runCommandGlobally(cmd *minicli.Command) chan minicli.Responses {
 	// Keep the original CLI input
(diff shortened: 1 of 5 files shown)


djfritz:
There are still issues with internal commands in web:

On the node running a webserver:

    mesh send all .record false vm info

Shows up in the history.

On a node that is accessed by another running a webserver:

    vm screenshot 0 140

Shows up in the history.

djfritz:
Everything else looks good.

djfritz:
viz command is broken in this:

Looks like a field broken or missing in an sprintf or something similar. Output includes:

"ccc15:vm-73:73:[]:[]" -- %!s(int=425)

djfritz:
That last one may be a bug, but the issue was on https://github.com/jcrussell/minimega/blob/bf34afc3edbc2c49b3f21c3f91ec45f7641faac1/src/minimega/dot.go#L82

djfritz:
And likely https://github.com/jcrussell/minimega/blob/bf34afc3edbc2c49b3f21c3f91ec45f7641faac1/src/minimega/dot.go#L89

djfritz:
Yes actually - both of those.

djfritz:
Jon is out of town, so I'm merging this and fixing the rest.

PR #230, opened 2015-06-05
python: better versioning

devinc:
Generated API files now contain minimega version, per @bjwrigh's request.

### DIFF
--- a/misc/python/genapi.py
+++ b/misc/python/genapi.py
@@ -41,7 +41,7 @@ CMD_BLACKLIST = [
 ]
 
 # API Version:
-VERSION = '2.0a1'
+VERSION = '2.0a2'
 # minimega version (set this before rendering)
 MM_VERSION = 'UNKNOWN'
 
@@ -133,7 +133,7 @@ if __name__ == '__main__':
     mm_bin = sys.argv[1]
     cmds = json.loads(subprocess.check_output([mm_bin, '-cli']))
     version_str = subprocess.check_output([mm_bin, '--version'])
-    MM_VERSION = version_str.split(None, 2)[1]
+    MM_VERSION = version_str.splitlines()[0]
 
     sys.stdout.write(render(cmds))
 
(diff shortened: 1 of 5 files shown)


misc/python/genapi.py
```diff
@@ -40,6 +40,11 @@
     #'mesh send',
 ]
 
+# API Version:
+VERSION = '2.0a1'
+# minimega version (set this before rendering)
+MM_VERSION = 'UNKNOWN'
```
djfritz: See https://github.com/sandia-minimega/minimega/blob/master/build.bash#L8

We auto-generate the version as the commit hash, and hand-tool a version number on releases. Could we do that here as well?

devinc: We could, but then the version produced in the API file when run with
different versions of the minimega binary would be incorrect. This way you
can always run the latest genapi.py with whatever version of minimega you
want bindings for.

djfritz: I'm confused - is the goal to have the generated binding warn/nag if there's a version mismatch? If so, wouldn't you want per commit matching, since we currently do have breaking changes pretty much weekly?

devinc: You get that functionality, plus the ability to generate compliant versions
of the minimega.py file that include any fixes to the binding template down
the line.

Actually, I probably also should have bumped the API version to 2.0a2 when
we made the last change to it.

djfritz:
It should be revision specific. See `minimega -version`. In fact, the genapi script should just invoke the version flag to write the version string with date to the generated script.

Also, let's add the genapi stuff to the build script here too.

devinc:
That is in fact what it does. It would probably be useful to use the whole
version string; right now I just use the git hash.

There are two different version numbers. There's the API version, and the
protocol version. It's probably confusing. If we don't care at all about
API versioning we can remove that one.

I'll add it to the build script.

devinc:
ok, how does this look?

djfritz:
LGTM?

PR #240, opened 2015-07-09
Web-based network rendering

jfktrey:
Force-directed graphs!
Fast WebGL rendering!
Visualize your network!

PR #242, opened 2015-07-10
minimega: export for gob

jcrussell:
Renaming a bunch of stuff so that gob encoding works properly. Should resolve some of the issues with #240.


djfritz:
This looks sane. Is it tested? Should I just merge?

jcrussell:
It was mostly done with gorename and I did some small testing on two nodes. Should be fine to merge.

djfritz:
LGTM

PR #243, opened 2015-07-13
minimega: change VM Info to take single mask.

jcrussell:
Change interface so that info takes a single mask and returns a single
value. Cleans up some code.


djfritz:
LGTM

PR #245, opened 2015-07-13
minimega: couple bug fixes.

jcrussell:
Change code to populate and use a single slice rather than throw away a
bunch of values.


djfritz:
LGTM

PR #250, opened 2015-07-14
legal: make sure all licenses are in LICENSES/

jcrussell:
Copied LICENSE files from subdirs into top-level LICENSES/


djfritz:
LGTM

PR #256, opened 2015-07-23
minimega: vm cleanup.

jcrussell:
Made some changes to how the VM code works so that it's a little easier
to add other types of VMs. Adding some more documentation for newer
functionality.


djfritz:
So how does sorting in vm info work now?

jcrussell:
minicli sorts based on the string value in the first column. vmSorter has
been dead code since minicli landed.

djfritz:
Track. LGTM

PR #259, opened 2015-07-23
ranges: fixed off-by-one bug.

### LINKED ISSUE #258: minimega: vm start/stop <id> doesn't work
Calling `vm start` or `vm stop` on a VM with the wildcard or a name works. Calling with an ID does nothing. This goes back at least to merging the VM++ branch.

Please triage.

@jcrussell

jcrussell:
Works correctly for single character names/IDs. Fixes #258.


djfritz:
LGTM

PR #260, opened 2015-07-24
meshage: disallow mesh dial yourself.

### LINKED ISSUE #244: meshage: mesh dialing yourself shold not be allowed
It currently is, establishes a connection, and fails moments later. meshage survives this, but it's nonetheless invalid. Add a check.

jcrussell:
Fixes #244. Don't allow meshage to connect to itself.

### DIFF
--- a/src/meshage/node.go
+++ b/src/meshage/node.go
@@ -31,6 +31,7 @@ package meshage
 
 import (
 	"encoding/gob"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


djfritz:
This doesn't cover the myriad ways to dial yourself - "localhost", "127.0.0.1", "fe80::..."

This probably needs to generate a list of internal IPs using net.Interfaces(), and compare against a given IP to dial to, or the resolved IP of a given host.

OR - meshage as it's currently implemented will allow the local connection, but that connection will hit an error after the handshake (~10-20 seconds) and kill that connection. We *could* just leave this as is.

jcrussell:
It does cover localhost and 127.0.0.1. It checks what the node calls itself
in meshage against what it calls itself in meshage.

djfritz:
Woop, you're right. LGTM

PR #265, opened 2015-07-29
minimega: rework screenshotting

### LINKED ISSUE #252: minimega: additional option for 'vm screenshot'
The `vm screenshot` API has a hardcoded path of `screenshot.png` in a VM's runtime directory. This works well enough until you have an external tool using screenshots while someone is using the web service. The web service grabs screenshots at a small resolution, and this may clobber images needed by external tools. It's in effect a race.

The web service doesn't actually use the image data written to disk - image data is passed in the minicli response itself. This opens up two options:

1. Allow specifying a path for an image, and have the web service point to /dev/null
2. Add an optional arg to the API to not write out an image.

I'm leaning towards 1. Thoughts?

@jcrussell

floren:
This is intended to fix #252. It maintains compatibility with the
old "vm screenshot" command while allowing you to say this:

	vm screenshot 0 file /dev/null
	vm screenshot 0 file /dev/null 100

Instead of writing out the screenshot file at the deepest level, I
pass back a slice of bytes containing the png data, then write
it out in vm_cli's cliVmScreenshot function. Previously, we were
writing to the file, then reading that back out into the response's
Body field, which made it fail when you give /dev/null as a file.

### DIFF
--- a/src/minimega/kvm.go
+++ b/src/minimega/kvm.go
@@ -283,26 +283,32 @@ func (vm *KvmVM) QueryMigrate() (string, float64, error) {
 	return status, completed, nil
 }
 
-func (vm *KvmVM) Screenshot(fpath string, size int) error {
+func (vm *KvmVM) Screenshot(size int) ([]byte, error) {
 	suffix := rand.New(rand.NewSource(time.Now().UnixNano())).Int31()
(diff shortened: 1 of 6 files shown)


floren:
@jcrussell @djfritz

djfritz:
LGTM

PR #266, opened 2015-07-30
Neoweb

jfktrey:
Better web interface, fixes for graph, button to reflow the graph, JSONificaton of host info

Everything is rolled into one page that automatically updates the data displayed (including host data).


djfritz:
From chrome's dev console:

    GET http://localhost:9001/hosts net::ERR_CONNECTION_REFUSED d3.min.js:1Nt.i.send d3.min.js:1i.(anonymous function) d3.min.js:1Nt d3.min.js:1(anonymous function) d3.min.js:1updateHosts glue.js:27

Is the 9001 port hardcoded somewhere?

jfktrey:
Yup, you're absolutely right. Replaced all instances of absolute paths with relative ones, so it should be fixed now.

djfritz:
LGTM

PR #274, opened 2015-08-11
minimega: fix #273.

### LINKED ISSUE #273: minimega: web reporting a lot of qmp errors
2015/08/06 22:20:09 ERROR web.go:170: write unix /tmp/minimega/395/qmp: broken pipe
    2015/08/06 22:20:09 http: multiple response.WriteHeader call

@jcrussell @jfktrey

Can you investigate? If it's not an issue, let's supress this error - there's a LOT of chatter with this error on big experiments.

jcrussell:
If the VM is not running, don't try to grab a screenshot from it.


djfritz:
LGTM

PR #275, opened 2015-08-12
Improved VM net command

### LINKED ISSUE #184: VM net disconnect does not update MM data
vm net disconnect 0 0 
vm info
(see that the interface still reports that the 0'th interface is still connected on the bridge,vlan,w/ mac, etc)

fritz has seen the bug if this isnt enough information to reproduce the bug

jcrussell:
When a network is disconnected, clear the name of the bridge. Split network connect/disconnect into VM code so that we can lock properly.

Closes #184.


djfritz:
LGTM

PR #276, opened 2015-08-12
minimega: refactored the launch code. Fixes #209.

### LINKED ISSUE #209: minimega: at least one dangling kvm process on entering VM_ERROR state
If you hit the race:

vm launch foo noblock
vm start foo

And get the VM into the error state with the 'qmp not ready' error, you'll notice that qemu for that VM is still running, because vm.start() didn't kill the process after putting it in the error state. Later, vm kill won't attempt to kill it because it's in the error state.

At least two things could happen here instead:

* Attempt to reissue commands on error VMs that aren't actually dead - like 'vm start'. That would require some more book keeping
* Always kill VMs when they enter the error state

jcrussell:
Lots of refactorings to properly use locks and hopefully eliminate the
chances of a race condition when calling `launch noblock` and then
immediately starting the VM.

### DIFF
--- a/src/minimega/command_socket.go
+++ b/src/minimega/command_socket.go
@@ -108,7 +108,7 @@ func readLocalCommand(dec *json.Decoder) (*minicli.Command, error) {
 }
 
 func sendLocalResp(enc *json.Encoder, resp minicli.Responses, more bool) error {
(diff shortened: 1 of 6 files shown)


djfritz:
LGTM, we'll push deeper teting of this in the 2.1 test window

PR #282, opened 2015-08-19
minimega.init: fix ordering for startup check

floren:
When starting minimega, it attempts to read the pid file, then checks
if that was successful. Unfortunately someone put a "sleep" between the
read and the check, meaning the check actually checks the return status of
"sleep". I fixed it and made the sleep longer because minimega sometimes
starts a little slowly.


floren:
@djfritz @jcrussell @jenareljam jeremy this is your script IIRC

djfritz:
UGHHHHHHHHHHH

Should we ditch the sleep and grab the PID ourselves instead of waiting on minimega to create a pidfile? Then we could just leve it in /var/run like everyone else (does posix talk about pid files?)

PR #291, opened 2015-08-24
minimega: fix concurrency bug.

jcrussell:
Turns out commands could run at the same time since we dropped the
command lock before the command actually completed. Fixed now.


djfritz:
LGTM

PR #295, opened 2015-08-24
minimega: centralize minicli handler registration

jcrussell:
Move all the minicli handler registration out of init functions and into
a single function. This allows us to use the logging framework which
isn't available until after logSetup.


djfritz:
LGTM

PR #301, opened 2015-08-28
Web adjustments

floren:
Change "machine list" to "vm list" and add fields for taps and IP4 IP6 addrs.

Also move the navbar to the top.


djfritz:
LGTM

PR #311, opened 2015-09-01
legal: add missing licenses.

jcrussell:


jcrussell:
@floren could have said LGTM... who can know for sure.

PR #312, opened 2015-09-01
minimega: move up logSetup and cliSetup.

jcrussell:
Needs to happen before -cli flag is processed.


jcrussell:
@devinc says LGTM.

PR #313, opened 2015-09-01
minimega: call logSetup and cliSetup before tests.

### LINKED ISSUE #310: tests: vm inject
Getting an invalid command on the build:

    --- FAIL: TestParseInject (0.00s)
            qcow_test.go:24: invalid command: vm inject src src.qc2 "my file":"Program Files/my file"

Please triage.

@jcrussell

jcrussell:
Need to register handlers before trying to run tests. Fixes #310.

### DIFF
--- a/src/minimega/init_test.go
+++ b/src/minimega/init_test.go
@@ -1,0 +1,13 @@
+// Copyright (2015) Sandia Corporation.
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+
+package main
+
+func init() {
+	// Setup up default logger to log to stdout at the debug level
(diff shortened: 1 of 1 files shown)


djfritz:
This will probably break the containers branch:

https://github.com/djfritz/minimega/blob/containers/src/minimega/main.go#L73

containerShim() does a different kind of log setup by dup2'ing an inherited fd from the calling minimega. To work around this, the shim could undo what logSetup does, but that seems like a hack.

jcrussell:
It won't affect the shim -- it's only used for running tests.

djfritz:
Woops. I didn't look at the filename and inferred that you moved the log setup into init() in minimega's main.go. I am a potato.

djfritz:
LGTM

PR #314, opened 2015-09-02
Vendorize me

devinc:


floren:
The pull request that adds 12,000 lines of Python... this is gonna wreck our ideological purity :)

devinc:
blame @djfritz... I was fine leaving it out

jcrussell:
Our go purity is only 78.6%, need to drop that 18.5% JavaScript.

bjwrigh:
most of the JS is the novnc stuff.... just saying.

djfritz:
Long discussion with Vince, other things coming, punting this.

PR #333, opened 2015-10-16
minimega: refactored `vm inject` to `disk`.

### LINKED ISSUE #332: minimega: change vm inject api
Currently the `vm inject` API creates a new qcow2 image with a backing file, mounts the new image, and copies files into it. There are two other use cases:

1. inject files into the base image without creating a new derivative image
2. just create a derivative image with no file copying

Perhaps we move `vm inject` into a new `disk` API:

    disk snapshot <src image> [dst image]
    disk inject <image> <files like /path/to/src:/path/to/dst>...

This would force the current `vm inject` API to be accomplished with two commands, but it enables the other use cases. If the current `vm inject` use case is common enough, we could add a composite API:

    disk inject  <src image> <dst image> <files like /path/to/src:/path/to/dst>...

Thoughts?

@jcrussell @bjwrigh @floren

jcrussell:
Closes #332. Split `vm inject` API into `disk snapshot` and `disk
inject`. Tested with:

qemu-img create foo.qc2 2G
qemu-nbd -c /dev/nbd0 foo.qc2
parted /dev/nbd0 mklabel msdos unit GB mkpart primary fat32 0 1 mkpart primary fat32 1 2
mkfs.fat /dev/nbd0p1
mkfs.fat /dev/nbd0p2
qemu-nbd -d /dev/nbd0
minimega$ disk snapshot foo.qc2 foo.qc2
minimega$ disk inject /tmp/minimega/files/foo.qc2 /bin/bash:/bin/bash
minimega$ disk inject /tmp/minimega/files/foo.qc2:2 /bin/bash:/bin/bash


djfritz:
LGTM

PR #334, opened 2015-10-19
web: add VNC links to the network graph

floren:
Wherever you can look at info for a VM, it's likely that you'll want to interact directly with that VM. Now, in the network graph, when you click a VM it will add a VNC


floren:
@djfritz @jcrussell

djfritz:
LGTM, though this whole enchilada may change.

Issue #315, opened 2015-09-04
New web interface crashes Firefox and other issues with usability


mkunz7:
Steps to reproduce:
1. Start minimega
2. Launch 30 VM's (less will probably cause the issue as well)
3. Open the web interface in firefox
4. Select more than 6 VMs to be displayed 

I made video
http://ku.nz/proof.zip
download, extract and drag into a browser to play

Other issues I think we should consider changing
-> Why can't I click on a VM in the VM List to launch no vnc?
-> Why isn't there a disk column?
-> Why can't I click on the VNC screenshot to launch the VM? Could we just rename it VNC Sessions and get rid of the connect button?
-> Sometimes when I click on Connect it doesn't open a new tab for novnc and I have to click it twice or three times
-> Sorting by IP address doesn't work
-> Can Network Graph be moved to a different tab? I think it really clutters the information

djfritz:
@jfktrey

rkaack:
Confirmed as well on OSX with Firefox. Memory grows and Firefox becomes unusable. Will provide screenshot if needed.

jfktrey:
Firefox issue makes sense. The way that new data is automatically pulled and integrated into the dataTable isn't optimized, so that's likely the root cause. FF tends to leak memory more readily.

Other issues are definitely noted and will be addressed. Some should go away as the different widgets become more integrated. Network Graph should be able to be moved easily.

djfritz:
Hold off on moving things until it's discussed here.

djfritz:
@jfktrey make this the priority

mkunz7:
It's really annoying to have multiple vm-#'s and have to highlight over each to find out which node it is on.

Can we either include a hostname next to the Screenshot Name? Or a link from VM List to launch NoVNC? Or both?

jcrussell:
@mkunz7 -- we just merged the PR that addressed some of these concerns. Thoughts?

mkunz7:
Skiing in Colorado right now. I'm excited to check it out when I get back.

mkunz7:
VNC no longer works in firefox. 

After clicking connect to a vnc session.  I'm getting either a 
```Server disconnected (code: 1006)```
or a 
```noVNC ready: native WebSockets, canvas rendering```

Memory usage still rises and doesn't clean up when I right click connect and open a new tab and then close it. 

Seems to be working fine in chrome. Admittedly haven't given things a through run through. Are we planning on addressing some of the other ui issues mentioned above?

[...]

Issue #298, opened 2015-08-26
minimega: don't display vnc 'connect' button on non-kvm VMs


djfritz:
see summary.

@jfktrey
@jcrussell

djfritz:
This is part of the containers push.

jcrussell:
Fixed by #453.

PR #338, opened 2015-10-21
Fixes for #298 and #315

jfktrey:
Fix for #298 : Connect button is no longer shown for non-kvm machines
Adds a bit of documentation to glue.js
Stylesheet tweaks
Speedups for the Firefox performance issues mentioned in #315 (base64-encoded images are now generated server-side)

### DIFF
--- a/misc/web/css/minimega.css
+++ b/misc/web/css/minimega.css
@@ -7,16 +7,14 @@ body.modal-open {
 }
 
 #logo {
-	float: left;
-	margin: 5px;
 	width: 200px;
+	margin: 0 22px;
+	position: relative;
+	top: -5px;
 }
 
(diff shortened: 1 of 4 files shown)


djfritz:
finally getting to this...

jcrussell:
This actually hides the connect buttons for Chrome/Firefox:

```
-        if (vm.type != "kvm") toAppend.find(".connect-vm-button").attr("visibility", "hidden");
+        if (vm.type != "kvm") toAppend.find(".connect-vm-button").css("visibility", "hidden");
```

jcrussell:
Tested with 30+ VMs on FF, seems okay. Going to merge and then apply the above patch.

PR #343, opened 2015-10-27
Well, schedule me impressed

jcrussell:
*Please don't merge this until the 2.3 window.*

@djfritz @floren @bjwrigh 

This adds a rudimentary scheduler to minimega. The overarching goal has been to *not* break existing minimega scripts -- you should be able to add multiple hosts to a namespace and launch VMs across them as if you were running everything on a single host. We've stuck to that pretty closely.

The scheduler itself is pretty dumb -- it schedules VMs using a N/M approach (where N is the number of VMs to launch and M is the number of machines) -- but the surrounding infrastructure seems to be working well. I think we can drop in smarter scheduling logic as separate issues/pull requests.

A quick walkthrough:

 * New commands: `namespace` and `nsmod`. `namespace` controls the currently active namespace -- allowing the user to enter a new namespace with `namespace <namespace>` or to run commands within a namespace `namespace <namespace> (cmd)`. It also allows you to inspect the hosts that belong to the namespace and see the status of currently running schedules. `nsmod` allows you to add or remove hosts from the namespace. Eventually, it should be used to tweak how we schedule VMs.
 * Modified behavior for `vm launch`. When you launch VMs and a namespace is active, the VMs are *not* launched immediately. Instead, they are put into a queue of VMs to schedule which is flushed when the user invokes `vm launch` with no additional arguments.
 * Modified behavior for `vm info`, `host`, `vm flush`. When a namespace is active, these commands are broadcast to all hosts that are part of the namespace and the results are collected on the issuing node. Equivalent to doing `mesh send <hosts> vm info` where `<hosts>` is the lists of hosts in the namespace.
 * Modified behavior for `vm start`, `vm pause`, `vm kill`, `vm hotplug`, `vm qmp`, `vm screenshot`, `vm tag`, `vm cdrom`. When a namespace is active, these commands now apply to VMs across the namespace on any host. minimega now tries to enforce globally unique VM names so these commands should use a name, rather than an ID, for the target. `vm kill 0` **will** kill all VMs with ID 0 on hosts that belong to the active namespace.
 * Modified behavior for `mesh send all`. When a namespace is active, `all` expands to all hosts in the current namespace -- not all nodes that are connected via meshage.
 * Modified behavior for `cc`. When namespace is active, there is an implicit filter for VMs running in the active namespace. To support this, we register a VM's namespace with the ron server so that we can filter the commands to send the client.
 * New command: `vm config cpu`. Configures the `-cpu` flag for qemu.
 * VLAN aliases. You can now create aliases for VLANs (e.g. `vm config net DMZ`). These aliases are namespace aware (so DMZ in namespace foo is different than DMZ in namespace bar). There is a new command, `vlans`, to inspect aliased VLANs. You can manually specify VLAN aliases with `vlans add <alias> <vlan>`.
 * Renamed `namespace` flag to `nameshpace` to avoid confusion.

Thoughts/concerns/why did you do that? Once CCC is back up, I'm planning to run some larger scale tests to see if anything breaks.


djfritz:
@jcrussell can you expand on `vm config namespace` and `vm config cpu`?

djfritz:
LGTM just doesn't do this justice

djfritz:
In the trembling grey of a spring dawn, when the birds were whispering in mysterious cadence among the trees, have you not felt that they were in wait for this merge? Where better than in this merge, sweet in its untold bugs, can we image the unfolding of a new minimega? The primeval man in offering the first commit to this branch is thereby transcended. He became human in thus rising above the crude necessities of mesh prefixes. He entered the realm of art when he perceived the subtle use of the useless.

In joy or sadness, minimega is our constant friend. We eat, drink, sing, and dance with it. We wed and christen with it. We dare not die without it. We have worshiped with debug logging, we have meditated over Jon's refactors, we have charged in battle array with the wildcard - ALL! We have even attempted to speak in the language of minimega. How could we live without it? It frightens one to conceive of a world bereft of its presence. What solace does minimega not bring to the bedside of the sick, what a light of bliss to the darkness of weary spirits? Its serene tenderness restores to us our waning confidence in the universe. When we are laid low in the dust it is minimega who lingers in sorrow over our graves.

LGTM

jcrussell:
Actually, I've been thinking about refactoring this a bit... ;)

Issue #345, opened 2015-11-02
Enhance vnc replay


floren:
We should extend the vnc playback parser so that in addition to keyboard and mouse events, we can have "load file" events in a VNC file. This will allow us to create a file containing a series of separate VNC playback files which will be played in sequence on a given VM, e.g.:

    0 LoadFileEvent,login.vnc
    1112233 LoadFileEvent,openbrowser.vnc
    1144000 LoadFileEvent,browseslashdot.vnc

djfritz:
@jcrussell this is a slick idea.

jcrussell:
Nice idea! Should be pretty simple to implement. Are you implementing it recursively? Hopefully people don't create `busy beaver` scripts:

```
$ cat busybeaver.vnc
0 PointerEvent,mask,x0,y0
100 PointerEvent,mask,x1,y2
100 LoadFileEvent,busybeaver.vnc
```

Perhaps we should have an `ExecFileEvent` which behaves like `exec` -- replacing the currently executing vnc playback with a new input file.

jcrussell:
@djfritz -- think we need `ExecFile` or should we just assume people won't want this behavior?

PR #347, opened 2015-11-03
minimega: allow scripted multi-file VNC playback

floren:
In addition to the existing PointerEvent and KeyEvent in VNC playback,
files can now include a LoadFile directive. This will cause the specified
file to be loaded and played. The format follows the usual VNC playback
format, e.g.:

	0:LoadFile,browse_slashdot.vnc
	10000000000:LoadFile,/home/john/recordings/reboot_windows.vnc

The filenames may be relative to the location of the current file,
or they can be absolute. Timestamps are honored as usual.

### DIFF
--- a/src/minimega/vnc.go
+++ b/src/minimega/vnc.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	log "minilog"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


src/minimega/vnc.go
```diff
@@ -234,10 +237,42 @@ func (v *vncKBPlayback) Run() {
 			v.err = res.Write(v.Conn)
 		} else if res, err := vnc.ParsePointerEvent(s[1]); err == nil {
 			v.err = res.Write(v.Conn)
+		} else if file, err := ParseLoadFileEvent(s[1]); err == nil {
+			if !filepath.IsAbs(file) {
+				// Our file is in the same directory as the parent
+				file = filepath.Join(filepath.Dir(v.file.Name()), file)
+			}
+			log.Warn("loading file %v", file)
```
jcrussell: Probably should change to log.Debug.

jcrussell:
Otherwise, LGTM.

floren:
Fixes #345

djfritz:
fixes #347

PR #351, opened 2015-11-04
igor/minidoc: fixing bad filepath joins.

### LINKED ISSUE #349: container and kvm instancePath is broken
https://github.com/sandia-minimega/minimega/commit/a86d92c025119db8d99911e19172982a3ad3b7a3#diff-48918408dbc4b874beb46309af5b0841L171

correctly fixed instancePath to use filepath.Join, which results in a path with no trailing "/"

A lot of things used code like:

    writeOrDie(vm.instancePath+"config", vm.Config().String())

Which now results in instance directory structures being flattened to:

    drwx------ 2 root root 120 Nov  3 23:47 0
    srwxr-xr-x 1 root root   0 Nov  3 23:47 0cc
    -rw-r--r-- 1 root root   8 Nov  3 23:47 0state
    drwx------ 4 root root  80 Nov  3 23:47 1
    -rw-r--r-- 1 root root 129 Nov  3 23:47 1config
    -rw-r--r-- 1 root root   4 Nov  3 23:47 1name
    -rw-r--r-- 1 root root   5 Nov  3 23:47 1state
    drwxr-xr-x 3 root root 160 Nov  3 23:39 files
    srwxr-xr-x 1 root root   0 Nov  3 23:47 minimega
    -rw-r--r-- 1 root root   4 Nov  3 23:47 minimega.pid

This shouldn't have happened. Poor original code and poor testing when parts of this were fixed.

Please triage. We need to go through all code to fix places that aren't using filepath.Join and fix it. String concatenation here is nuts.

@jcrussell @floren

jcrussell:
Use filepath.Join rather than string concat. Fixes #349.

### DIFF
--- a/src/igor/sub.go
+++ b/src/igor/sub.go
@@ -148,27 +148,30 @@ func runSub(cmd *Command, args []string) {
 	}
 
 	// make kernel copy
-	kdest, err := os.Create(igorConfig.TFTPRoot + "/igor/" + subR + "-kernel")
+	fname := filepath.Join(igorConfig.TFTPRoot, "igor", subR+"-kernel")
+	kdest, err := os.Create(fname)
 	if err != nil {
-		fatalf("%v", err)
+		fatalf("failed to create %v -- %v", fname, err)
 	}
 	io.Copy(kdest, ksource)
 	kdest.Close()
 	ksource.Close()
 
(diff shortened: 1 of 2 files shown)


djfritz:
LGTM

PR #360, opened 2015-11-06
minimega: read took a chill pill

### LINKED ISSUE #358: minimega: aggressive locking breaks the web interface
https://github.com/sandia-minimega/minimega/commit/c506e0bdaa3ce0e8eaeb49d315d2e22d6ad7cc83

has the side effect that when you do something like `read foo`, and `foo` is a long script (taking minutes to complete, which isn't uncommon), the web interface is effectively frozen. 

The web interface backend commands need to safely ignore the lock.

@jcrussell

jcrussell:
Read no longer acquires the command lock for its whole duration -- it
releases and aquires it for each command that it runs. This allows the
web interface (and other interfaces to minimega) to continue to work
even when we are reading a long file.

Fixes #358.

### DIFF
--- a/src/minimega/misc_cli.go
+++ b/src/minimega/misc_cli.go
@@ -121,6 +121,13 @@ func cliRead(c *minicli.Command, respChan chan minicli.Responses) {
 	}
 	defer file.Close()
 
+	// HACK: We *don't* want long-running read commands to cause all other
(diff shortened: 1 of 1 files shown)


floren:
I dig it

djfritz:
LGTM - tested with ~1000 line script, 18 nodes, and the web interface. Appears to work fine.

PR #364, opened 2015-11-10
minimega: fix #359.

### LINKED ISSUE #359: minimega: vnc api still cannot match on vm name
see summary, we're forced to use the vm id still. I thought we fixed this already, so maybe I'm just being dumb somewhere.

@jcrussell @floren

jcrussell:
Whoops, wrong arg to format string. Should fix #359, needs testing but someone took all the nodes.


djfritz:
tested and LGTM

jcrussell:
Thanks

PR #366, opened 2015-11-10
minimega: fix #365.

### LINKED ISSUE #365: minimega: disk_test.go
```
--- FAIL: TestParseInject (0.00s)
        disk_test.go:23: invalid command: disk inject image.qc2 "my file":"Program Files/my file"
```

We should just get rid of this test, it was useful when we had custom parsing but isn't really useful now.

jcrussell:
Remove test file that was useful when we had lots of custom parsing --
not really useful now that we have minicli.


djfritz:
LGTM

PR #369, opened 2015-11-17
minimega: remove link local from VM IPv6 column.

jcrussell:
Filter out the IPv6 link local address for `vm info`.


djfritz:
LGTM even though you are a debug statement black hole.

PR #371, opened 2015-11-20
Mouse Movement Tracking

mkunz7:
Update the noVNC status bar to include Mouse Coordinates

### DIFF
--- a/misc/web/novnc/include/input.js
+++ b/misc/web/novnc/include/input.js
@@ -303,6 +303,12 @@ var Keyboard, Mouse;
             var evt = (e ? e : window.event);
             var pos = Util.getEventPosition(e, this._target, this._scale);
             if (this._onMouseMove) {
+                if (noVNC_status.innerHTML.indexOf('|') == -1){
+                    noVNC_status.innerHTML += " | X:"+pos.x+" Y:"+pos.y;
+                }
+                else{
+                    noVNC_status.innerHTML = noVNC_status.innerHTML.split('|')[0] + "| X:"+pos.x+" Y:"+pos.y;
+                }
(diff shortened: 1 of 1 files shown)


djfritz:
@jcrussell @floren 

This is vendorized code which we don't maintain. If we did this we'd have to keep a patch somewhere in the repo to apply when we periodically upgrade to the latest novnc release. 

What's the need for this?

mkunz7:
I needed an easier way to make modifications to VNC scripts. I've been creating VNC scripts that do X task taking X variables and when writing these scripts by hand or fixing them I've needed the exact mouse coordinates. 

Before this PR, you would need to start a recording via the cli, move the mouse, stop the recording in the cli, open the save file in a text editor, and scroll to find the mouse coordinates you care about. Now I can just open VNC and see exactly where the new coordinate I care about is at.

I'm not the one to determine the best way of integrating this into the project, but I believe there should be an easier process than what is currently used when writing VNC scripts. With this being such a minimal change to vendor code that rarely updates, I figured these 6 lines can add a great value. But ultimately it's what's best for the project.

jcrussell:
Would it be easier to have this be a separate JS file that we include that adds another event listener for the mouse move event? Then we would just inject the extra script (perhaps `misc/web/js/coords.js`) on the noVNC page. We would probably want to leave the noVNC status bar alone and add it somewhere else on the page to make it the most portable.

mkunz7:
Maybe we could add a named empty div object to the html page to store the coordinates. Im thinking next to the control alt delete button. A checkbox to show mouse coordinates when checked. And a new JavaScript event handler in another JavaScript file to check if the box is checked and if so update the page.

PR #372, opened 2015-12-01
update init script to launch minimega with the correct working directory

jenareljam:
This makes commands that assume the CWD is $MINIMEGA_DIR (such as the "web" command) work when minimega is run as a daemon.

### DIFF
--- a/misc/daemon/minimega.init
+++ b/misc/daemon/minimega.init
@@ -82,7 +82,9 @@ start() {
     return 1
   fi
   mkdir -p $MM_RUN_PATH
+	pushd $MINIMEGA_DIR &>/dev/null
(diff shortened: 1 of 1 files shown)


floren:
LGTM

jcrussell:
Ouch, tabs vs spaces ><

jenareljam:
Ouch. What the heck? My .vimrc is set to use spaces for bash scripts... gr...

jenareljam:
D'oh. Looks like it's not. Must have happened with the recent change. Man I miss gofmt in every other language...

jcrussell:
@jenareljam At least there's one for python now: https://github.com/google/yapf

jenareljam:
Nice find! Looks like it has a vim plugin too: https://github.com/google/yapf/blob/master/plugins/yapf.vim

PR #376, opened 2015-12-16
Code improvements and fix #374

### LINKED ISSUE #374: tabular width mismatch on minidoc
If I run "vm info" from within minidoc, I get a tabular mismatch error from minicli/responses.go, the validTabular function. I added a little more debugging, here's what I see:

tabular width mismatch, host: freyr, row: 0. len(header) = 23, len(v.Tabular[0] = 24
header = [id name state memory vcpus type vlan bridge tap mac ip ip6 bandwidth migrate disk snapshot initrd kernel cdrom append uuid cc_active tags], v.Tabular[j] = [freyr 0 vm-0 BUILDING 2048 1 kvm [] [] [] [] [] [] []  [generic_vm.qcow2] true     ada7cdd4-fd7a-4614-8db5-c78257741005 false map[]]

Basically, the header doesn't contain a "host" field, but the Tabular entry has the host as the first field.

If I say .annotate false, it works fine.

jcrussell:
Update: includes fix for issue #374.

Split paging from miniclient into separate package so that we don't have
to import unsafe in miniclient. This allows us to remove the copied code
from minidoc.

@floren


djfritz:
LGTM

PR #381, opened 2016-01-07
minimega: fix #378

### LINKED ISSUE #378: minimega: `viz` command hangs
`viz` command hangs because it uses `globalVmInfo` which invokes `runCommandGlobally` that requires a lock.

jcrussell:
Stealing code from namespaces to make sure we don't try to grab the cmdLock if we already have it.


djfritz:
LGTM

Issue #395, opened 2016-01-11
minimega: race in kvm.go checkInterfaces


djfritz:
This could exist in the equivalent source in container.go, though it's implemented differently.

Stacktrace attached. ranging on another vm's vm.Networks while it's still being created by the other vm is racy.

[mm.log.txt](https://github.com/sandia-minimega/minimega/files/86400/mm.log.txt)

djfritz:
@jcrussell

Issue #391, opened 2016-01-11
minimega: vm.setState race condition


jcrussell:
We currently call `vm.setState` in launch from a deferred function because it needs to acquire the VM lock after launch or whatever releases it. This creates a race condition as another goroutine might also be waiting on the VM lock. We should make it so that `vm.setState` assumes that the caller already holds the lock. Perhaps we should also add a new function to BaseVM:

```
func (vm *BaseVM) setError(err error) {
    vm.tags["error"] = err.String()
    vm.setState(VM_ERROR)
}
```

This would be useful for debugging while VMs are in the error state without having to hunt through the logs.

djfritz:
I'm for this. Doing this will resolve #388

djfritz:
I like the error tag as well.

jcrussell:
Cool -- I'm on it.

PR #398, opened 2016-01-11
minimega: vm refactoring.

### LINKED ISSUE #388: minimega: launching containers hangs
If a container launch() has an error after the containerShim has launched, launch() will never return as a call to setState occurs while the vm lock is held. Need to refactor launch() in container.go.

@jcrussell @floren

jcrussell:
Cleans up a lot of code and fixes #388, #391, #395.

Only tested with `minitest` for sanity, a larger test would be beneficial before merging.


src/minimega/kvm.go
```diff
@@ -520,54 +490,48 @@ func (vm *KvmVM) launch(ack chan int) (err error) {
...
 	go vm.asyncLogger()
 
-	ack <- vm.ID
-
 	// connect cc
 	ccPath := filepath.Join(vm.instancePath, "cc")
-	err = ccNode.DialSerial(ccPath)
-	if err != nil {
+	if err := ccNode.DialSerial(ccPath); err != nil {
+		// TODO: Should we kill the VM?
```
djfritz: No, we should simply warn on the error. There are scenarios where people will use the qemu override to disable the cc virtio port for various reasons. Not common enough to justify making the cc port a tunable, but just the same it shouldn't prevent folks from still running a vm.

src/minimega/container.go
```diff
@@ -574,26 +574,14 @@ func NewContainer(name string) *ContainerVM {
 }
 
 func (vm *ContainerVM) Launch(ack chan int) error {
-	go vm.launch(ack)
+	vm.asyncLaunch(ack)
```
djfritz: The call stack becomes

*ContainerVM.Launch() -> *ContainerVM.asyncLaunch() -> *ContainerVM.launch()

The top two launch wrappers don't appear to do much of anything, can this be collapsed here and in kvm or is there a reason to have 3 layers of launch indirection that I'm missing?

jcrussell: Leaving one layer of indirection -- `*ContainerVM.launch()` is called via `*ContainerVM.Launch` and `*ContainerVM.Start` so it's a bit cleaner to have the lock acquired outside of `*ContainerVM.launch()`.

jcrussell:
ContainerVM.Start and KvmVM.Start are now almost exactly the same except that one calls `vm.q.Start` and the other calls `vm.thaw`. Stop is similar. Launch is identical.

It's not a lot of duplicated code but it might be worth thinking about how to lift these into BaseVM. Maybe that just complicates things later. Potentially, BaseVM could have function pointers for the low level start/stop/launch functions that the ContainerVM and KvmVM populate in their respective New functions.

Thoughts?

djfritz:
I'd prefer not to elevate these. There's a scenario where we have lots of vm types moving forward. I think the current vm interface is sufficient. There's also enough unknowns with the new mobile stuff and other big ticket items that I'd hold of on refactoring other stuff until we have a clear picture of where things need to go.

src/minimega/vmlist.go
```diff
@@ -190,7 +190,9 @@ func (vms VMs) launch(name string, vmType VMType, ack chan int) error {
 	vms[vm.GetID()] = vm
 	vmLock.Unlock()
 
-	return vm.Launch(ack)
+	go vm.Launch(ack)
```
djfritz: This was broken before and is continued here. We should be able to get an error back to minicli on a launch error if we're not using noblock. In this design and in the previous design launch errors just vanish unless you're on a local terminal and see the error log come out on stderr.

The comment above this function also states that the function should block until the launch is complete. That's no longer true here, though it probably should be, with the specific implementation's Launch() returning and error, and concurrency/blocking occuring in the cli/upper level call (which is half-does now).

djfritz:
See the above comment.

https://github.com/sandia-minimega/minimega/blob/master/src/minimega/vm_cli.go#L1191

The code in vm_cli.go has some quasi-dead code (launch only returns errors on duplicate names). 

It shouldn't be this convoluted to start a vm. We're doing something wrong.

jcrussell:
Working on a few more small tweaks, don't merge just yet.

jcrussell:
Alright, I'm done making tweaks.

djfritz:
@floren you satisfied? LGTM otherwise.

floren:
Yep!

PR #399, opened 2016-01-12
minimega: fix deadlock in processCommands.

### LINKED ISSUE #394: minimega: regression in cli
https://github.com/sandia-minimega/minimega/commit/3e06eda8f53aaeac80facd0016c56c707d590070

Created a regression in at least one spot:

https://github.com/sandia-minimega/minimega/blob/master/src/minimega/cli.go#L86

If cliPreprocessor returns an error, the function deadlocks on sending into a buffered channel that it forward()s later in the function.

Please triage.

@jcrussell

jcrussell:
Fixes #394. Make sure to send from a goroutine to avoid deadlock.


djfritz:
LGTM

PR #401, opened 2016-01-14
protonuke: properly fill the httpImage

floren:
len(m.Pix) is actually 1024*768*4


floren:
@djfritz @jcrussell

jcrussell:
LGTM

PR #402, opened 2016-01-14
protonuke: allow user to specify size of image served by http/https

floren:
This should make it easier to serve a bigfile in instances when we want massive traffic.


floren:
@jcrussell @djfritz

djfritz:
So if we ask for a 1G image, protonuke will generate a random one on every request?

floren:
Actually no. If you look at the code, you'll see that httpSetup() is the only function which calls httpMakeImage(), and that creates a variable called httpImage which is used whenever someone hits /image.png.

If protonuke once generated a random image for each request, it doesn't any more.

floren:
I agree that generating a random 1G image might be a bit much if we do it for every request. If we want to generate a fresh image each time (important for caching), then I will withdraw this PR and submit a new one which makes protonuke generate a new image each time.

src/protonuke/protonuke.go
```diff
@@ -34,6 +34,7 @@ var (
 	f_v4       = flag.Bool("ipv4", true, "use IPv4. Can be used together with -ipv6")
 	f_v6       = flag.Bool("ipv6", true, "use IPv6. Can be used together with -ipv4")
 	f_report   = flag.Duration("report", time.Duration(10*time.Second), "time between reports, set to 0 to disable")
+	f_imageSize = flag.Int("imagesize", 3, "size of image to serve in http/https pages")
```
djfritz: flag should be -httpimagesize or similar to match other protocol specific fields.

djfritz: s/size of image/size in mb of image/

djfritz:
LGTM

PR #409, opened 2016-01-25
minitunnel: fix race.

jcrussell:
There was a race in minitunnel -- if you forwarded a port and used it
immediately, before the TID was registered, then some of the data would get
lost causing the tests to hang. Moved spawning goroutine to after registering
the TID to fix. Ran tests for ten minutes with no issues.

Also, rewrote tests so that we don't call T.FailNow from separate goroutines as
per the [documentation](https://golang.org/pkg/testing/#T.FailNow).


djfritz:
tested, appears to work fine

LGTM

PR #410, opened 2016-01-27
minimega: fix noblock issues.

jcrussell:
Remove two races caused by `vm launch ... noblock`:

1. Need to create copy of `vmConfig` before running the next command (in case the next command changes `vmConfig`).
2. Need to block on `vm start all` (and other vm target commands) until all outstanding `vm launch ... noblock` commands have finished creating vms.

Note: this is a cleaner fix than the one that was pushed to @djfritz's fork.


djfritz:
testing this now

djfritz:
LGTM

PR #414, opened 2016-02-03
minimega: change deploy suffix.

jcrussell:
Rather than use a random suffix, append the unix timestamp. This allows
you to figure out what the latest binary is without checking `ls -l`.


djfritz:
LGTM

PR #416, opened 2016-02-03
Add /tilevnc to web interface

floren:
If you go to /tilevnc on the web interface, you can view all the VNC screens at once with no additional information. Works best on Chrome, Firefox uses a lot of CPU.


floren:
@djfritz

djfritz:
LGTM

PR #419, opened 2016-02-03
minicli: fix .filter

jcrussell:
Need to forward the filtered resps rather than the original resps...


floren:
LGTM

djfritz:
Such a big patch probably needs a longer review cycle.

PR #420, opened 2016-02-05
protonuke: add custom certificate flags (issue #418)

floren:
With the -httptlscert and -httptlskey flags, you can specify
your own certificate instead of having protonuke generate one.
Note that you must specify both.

@djfritz


djfritz:
LGTM

PR #422, opened 2016-02-17
protonuke: Add timeouts to http.

floren:
Without this, protonuke's http and https threads can end up hung on a read if something goes weird in the connection.


djfritz:
LGTM

PR #423, opened 2016-02-23
Per interface qos

cdshann:
Initial implementation of adding quality-of-service constraints on individual interfaces using traffic control (tc). Currently only adding packet loss and delay is supported. 

qos list
qos add \<interface\> loss \<loss %\>
qos add \<interface\> delay \<delay in microseconds\>
qos remove \<interface\>
qos remove all

The add command also supports chaining multiple qos parameters

qos add \<interface\> loss \<loss %\> delay \<delay in microseconds\>

This is implemented by embedding a *Qos field in each Tap struct. The Qos struct contains a map of qos parameters that are used to build qos commands. If there is no qos associated with a tap the *Qos pointer is set to nil.

Currently this implementation assumes that tap names are all unique on individual hosts. Otherwise the syntax would need to be changed to include bridge names when adding qos constraints on a given tap.

Comments? I am planning on improving the minicli patterns and handler structure.


jcrussell:
Some feedback:

* use `time.ParseDuration` for delay, use some sane default unit if no units
* don't stack commands, create `qos config` commands
* make `qosAddCmd` have `*qos` receiver
* need `qosTaps`? could store qos in Tap struct (would need to change `Bridge.Taps` to `map[string]*Tap`)

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"fmt"
+	"minicli"
+	"strconv"
+)
+
+var qosCLIHandlers = []minicli.Handler{
+	{
+		HelpShort: "add qos constraints to an interface",
+		HelpLong:  "Fritz Fritz Fritz",
```
djfritz: We can populate this before merging.

src/minimega/qos.go
```diff
@@ -0,0 +1,140 @@
...
+// Map of tap names to bridges
+var qosTaps map[string]*Bridge
+
+// Tap field enumerating qos parameters
+type Qos struct {
+	params map[string]string
+}
+
+func init() {
+	qosTaps = make(map[string]*Bridge)
```
djfritz: Instead of a new global, embed QoS parameters in the existing tap struct.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,75 @@
...
+		Patterns: []string{
+			"qos list",
+			"qos <add,> <interface> <loss,> <percent>",
+			"qos <add,> <interface> <delay,> <ms>",
+			"qos <add,> <interface> <loss,> <percent> <delay,> <ms>",
```
djfritz: It looks like the original reason for supporting stacking here is that tc requires all of the config to happen in a single call to tc. It looks like the implementation could support ditching the stacked solution here and instead allowing the user to issue 2 commands (loss, then delay) and have tc be called with any existing config on a given tap:

    qos add mega_tap0 loss 50%
    qos add mega_tap0 delay 50ms

now mega_tap0 should have both because the second command still knew in the qos.params map about the loss setting.

So I would get rid of the stacked API.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"minicli"
+	"strconv"
+)
+
+var qosCLIHandlers = []minicli.Handler{
+	{
+		HelpShort: "add qos constraints to an interface",
+		HelpLong:  "Fritz Fritz Fritz",
+		Patterns: []string{
```
djfritz: I'd update the patterns to support finding interfaces better and change some semantics:

    qos show [vm or interface]
    qos <add,> <vm> <interface index> <loss,> <percent>
    qos <add,> <vm> <interface index> <delay,> <ms>
    qos <add,> <interface> <loss,> <percent>
    qos <add,> <interface> <delay,> <ms>
    clear qos
    clear qos <vm> <interface index>
    clear qos <interface>

@jcrussell should comment

cdshann:
I am using qosTaps to avoid having to iterate over all bridges+taps when removing qos from all taps, but if this isn't a concern I will get rid of the global.

@djfritz is correct that the current implementation supports adding multiple constraints so I will remove the stacked API.

djfritz:
If iterating on bridges/taps ever becomes a real problem, we'll refactor the bridge code to have maps of maps or something. In practice we don't usually ever have more than a few thousand taps so iterating isn't very expensive.

jcrussell:
Replying to updated patterns:

When I talked with Craig about this, it sounded like we may want to have `qos config` commands that work similarly to `vm config`. `tc` doesn't allow you to change parameters once it's started so you have to kill the process and restart it each time you change the loss or delay (or any of the other parameters we eventually wrap). I think the patterns should be something similar to:

```
qos <config,> <loss,> <percent>
qos <config,> <delay,> <duration>
qos <start,> <vm> <interface index>
qos <start,> <interface>
qos
clear qos
clear qos config
clear qos config <loss,>
clear qos config <delay,>
clear qos <vm> <interface index>
clear qos <interface>
```

Note: use .filter if you want specific interface when just running `qos`.

djfritz:
A primary use case is updating qos parameters at runtime often - an example would be a device moving away from a wifi AP, and the loss % increasing according to a distance-square law. The API should support seamless kill/restart of tc already, and once it does, configuring in the way I mentioned isn't problematic.

jcrussell:
If that's the case, won't killing/restarting `tc` lead to weird spikes in the loss/delay since tc won't be running? It seems like we need a way to configure a running `tc` instance, if possible. Then, we definitely don't need the `qos config` commands.

djfritz:
Actually I'm confused. `tc` doesn't stay running, it applies whatever qdisc you supply it and exits. There is a 'change' operation on `tc` as well. Where do we have to kill and restart anything?

cdshann:
Looks like I am bad at reading man pages. You can alter tc contrains with the update command. I will update the code to support the example by @djfritz

djfritz:
tc qdisc add dev eth0 root netem delay 100ms

and then later i can change it

    tc qdisc change dev eth0 root netem delay 100ms 10ms

jcrussell:
Looking pretty good. For consistency with the other minimega commands, please update
```
qos <remove,> <interface>
```
to
```
clear qos [interface]
```

cdshann:
Updated per @jcrussell feedback. @djfritz what else does this need?

djfritz:
Looks like it just needs testing. I'll have some time this week to start tinkering with it. Stay tuned. 

This may also slip to the 2.3 window, assuming I ever stop finding little nits with the containers implementation...

djfritz:
The next issue we'll add around QoS is a global and network policy:

    qos add global loss <loss %>
    qos add network loss <network> <loss %>

These have to actually track taps coming and going from networks, and new taps that get created. It shouldn't be a huge lift though. 

Let's hold off on this for now though.

djfritz:
@cdshann 

I just noticed that this doesn't have a bandwidth limiter. Can we toss in support for the token bucket filter while we're at it?

    qos <add,> <interface> <bandwidth,> <maximum bandwidth>

See http://linux.die.net/man/8/tc-tbf

Bandwidth here is according to the tc rules, and could be specified like `200kbit`.

The `tc` command should be simple enough too (I think...):

    tc qdisc add dev <interface> root tbf rate <maximum>

cdshann:
We only need to agree on default values for the burst and latency parameters for the tbf qdisc. See http://linux.die.net/man/8/tc-tbf

jcrussell:
Also, change "all" -> Wildcard.

jcrussell:
Path to merging:

- [ ] Add `minitest` tests
- [ ] Document interaction with namespaces

jcrussell:
I've been thinking about how this will work with namespaces and I want to change the API. This is what we have currently:

```
    qos <add,> <interface> <loss,> <percent>
    qos <add,> <interface> <delay,> <duration>
    qos <add,> <interface> <rate,> <bw>
```

This is what I think it should be:

```
    qos <add,> <target> <interface> <loss,> <percent>
    qos <add,> <target> <interface> <delay,> <duration>
    qos <add,> <target> <interface> <rate,> <bw>
```

Where `<target>` has the same meaning as `vm start` and `<interface>` is the interface number. This allows the `qos` API to interact *much* better with namespaces as we'll automatically figure out the right machine/bridge/tap to run `tc` on. I think the change should be relatively minor but would greatly improve the usability.

@djfritz @cdshann @floren

jcrussell:
I'm leaning towards merging #501 *before* this and then porting this code to apply to the post-merge code base. I think we're reasonably close on #501 and it would be cleaner to have this be a patch to the bridge package.

cdshann:
@jcrussell and I talked about this yesterday and we are going to wait until #501 is merged as the qos functionality will need to be absorbed into the new bridge code anyways.

src/minimega/vmlist.go
```diff
@@ -539,6 +540,51 @@ func (vms VMs) CleanDirs() {
...
 
+func (vms VMs) ListQoS(resp *minicli.Response) {
+	vmLock.Lock()
+	defer vmLock.Unlock()
+
+	resp.Header = []string{"vm", "bridge", "tap", "max_bandwidth", "loss", "delay"}
+	var data [][]string
+
+	for _, vm := range vms {
+		if !inNamespace(vm) {
```
jcrussell: This isn't needed -- PR #554 moved this check to `vms.apply`.

src/bridge/qos.go
```diff
@@ -4,13 +4,21 @@ import (
...
 	"fmt"
 	log "minilog"
-	"strconv"
 )
 
 // Used to calulate burst rate for the token bucket filter
 const KERNEL_TIMER_FREQ uint64 = 250
 const MIN_BURST_SIZE uint64 = 2048
 
+type QosParams struct {
```
jcrussell: Shouldn't these be float64s? If you were using the zero value string to indicate no change, you'll probably have to make them *float64s.

src/bridge/tap.go
```diff
@@ -128,6 +128,8 @@ func (b *Bridge) destroyTap(t string) error {
 
 	tap.Defunct = true
 
+	b.clearQos(tap)
```
jcrussell: Check error?

src/minimega/qos_cli.go
```diff
@@ -40,8 +40,8 @@ Examples:
 		Patterns: []string{
 			"qos <add,> <target> <interface> <loss,> <percent>",
 			"qos <add,> <target> <interface> <delay,> <duration>",
-			"qos <add,> <target> <interface> <rate,> <bw>",
-		}, Call: wrapVMTargetCLI(cliQos),
+			"qos <add,> <target> <interface> <rate,> <kbit, mbit, gbit>",
```
jcrussell: Probably shouldn't have spaces between units -- I *think* minicli will trim them but I'm not certain.

src/minimega/vm.go
```diff
@@ -78,6 +79,11 @@ type VM interface {
 	// NetworkDisconnect updates the VM's config to reflect that the specified
 	// tap has been disconnected.
 	NetworkDisconnect(int) error
+
+	// Qos functions
+	GetQos() [][]string
```
jcrussell: Return []bridge.QosParams instead?

src/minimega/vm.go
```diff
@@ -419,6 +425,80 @@ func (vm *BaseVM) UpdateBW() {
...
+
+	br, err := getBridge(bName)
+	if err != nil {
+		return err
+	}
+
+	return br.UpdateQos(tapName, qosp)
+}
+
+func (vm *BaseVM) ClearAllQos() {
```
jcrussell: Return first error?

src/minimega/vm.go
```diff
@@ -419,6 +425,80 @@ func (vm *BaseVM) UpdateBW() {
...
+	if tap >= len(vm.Networks) {
+		return fmt.Errorf("invalid tap index specified: %d", tap)
+	}
+	nc := vm.Networks[tap]
+	b, err := getBridge(nc.Bridge)
+	if err != nil {
+		return err
+	}
+
+	b.ClearQos(nc.Tap)
```
jcrussell: Return this.

src/bridge/qos.go
cdshann: I agree, this was mostly to avoid reconverting back to a string before passing the arguments as part of a call to processWrapper() but that doesn't make sense outside qos being used through the command line interface.

src/minimega/vm.go
cdshann: Not sure how we would include the vm's network information (bridge name, tap name, etc) if this function returned []bridge.QosParams.

jcrussell: Moot, killing this in favor of adding to `vm info`.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,184 @@
...
+	if c.StringArgs["tap"] == Wildcard {
+		return fmt.Errorf("qos for wildcard taps not supported")
+	}
+
+	tap, err := strconv.ParseUint(c.StringArgs["tap"], 10, 32)
+	if err != nil {
+		return fmt.Errorf("invalid tap index %s", c.StringArgs["tap"])
+	}
+
+	if tap < 0 {
```
jcrussell: ParseUint => always false.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,184 @@
...
+		return fmt.Errorf("invalid tap index %d", tap)
+	}
+
+	return makeErrSlice(vms.ClearQoS(target, uint(tap)))
+}
+
+func cliUpdateQos(c *minicli.Command, resp *minicli.Response) error {
+
+	var err error
```
jcrussell: Not used.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,184 @@
...
+
+	} else {
+
+		qos.NetemParams = &bridge.NetemParams{}
+
+		// Drop packets randomly with probability = loss
+		if c.BoolArgs["loss"] {
+			loss := c.StringArgs["percent"]
+			v, err := strconv.ParseFloat(loss, 64)
+			if err != nil || v >= float64(100) {
```
jcrussell: v < 0?

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return err
+	}
+
+	return makeErrSlice(vms.UpdateQos(target, uint(tap), qos))
+
+}
+
+func cliParseQos(c *minicli.Command) (*bridge.Qos, error) {
+
```
jcrussell: I'd kill the newline after the `func....` line.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,184 @@
...
+	qos := &bridge.Qos{}
+
+	// Determine qdisc and set the parameters
+	if c.BoolArgs["rate"] {
+
```
jcrussell: And these newlines too.

src/minimega/vm.go
```diff
@@ -419,6 +426,80 @@ func (vm *BaseVM) UpdateBW() {
...
+
+	return nil
+}
+
+func (vm *BaseVM) GetQos() []*bridge.Qos {
```
jcrussell: Think we can kill this function.

src/minimega/vmlist.go
```diff
@@ -539,6 +540,43 @@ func (vms VMs) CleanDirs() {
...
+}
+
+func (vms VMs) ClearAllQos(target string) {
+	vmLock.Lock()
+	defer vmLock.Unlock()
+
+	// Clear qos for all vm taps
+	applyFunc := func(vm VM, wild bool) (bool, error) {
+		vm.ClearAllQos()
+		return true, nil
```
jcrussell: Shouldn't this be `return true, vm.ClearAllQos()`?

src/minimega/vm.go
```diff
@@ -419,6 +426,80 @@ func (vm *BaseVM) UpdateBW() {
...
+
+	br, err := getBridge(bName)
+	if err != nil {
+		return err
+	}
+
+	return br.UpdateQos(tapName, qos)
+}
+
+func (vm *BaseVM) ClearAllQos() {
```
jcrussell: Return first error.

src/bridge/qos.go
```diff
@@ -0,0 +1,143 @@
...
+	var qdisc string
+
+	if t.Qos == nil {
+		err := t.qosInitialize()
+		if err != nil {
+			return err
+		}
+	}
+
+	if qos.TbfParams != nil {
```
jcrussell: Hm, I know the current qos API won't update both fields at the same time but we may have some code that does so in the future. It would probably be best to check and run both if both are non-nil.

src/bridge/qos.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if t.Qos == nil {
+		return nil
+	}
+	return b.getQos(t)
+}
+
+// Return a copy of the tap's Qos struct
+func (b *Bridge) getQos(t *Tap) *Qos {
```
jcrussell: This should be a deep copy.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,174 @@
...
+		qos.TbfParams = &bridge.TbfParams{}
+
+		var unit string
+		var bps uint64
+		rate := c.StringArgs["bw"]
```
jcrussell: nit: drop unit var, rate += "kbit"...

src/minimega/vm.go
```diff
@@ -435,6 +442,79 @@ func (vm *BaseVM) UpdateBW() {
...
+	nc := vm.Networks[tap]
+	b, err := getBridge(nc.Bridge)
+	if err != nil {
+		return err
+	}
+
+	return b.ClearQos(nc.Tap)
+}
+
+func (vm *BaseVM) GetQos() []*bridge.Qos {
```
jcrussell: Kill this func?

jcrussell:
Code is coming along nicely, think we mainly need to test to make sure it works properly.

src/bridge/qos.go
```diff
@@ -0,0 +1,210 @@
...
+	log.Info("clearing qos for tap %s", tap)
+
+	t, ok := b.taps[tap]
+	if !ok {
+		return fmt.Errorf("tap %s not found", tap)
+	}
+	return b.clearQos(t)
+}
+
+func (b *Bridge) clearQos(t *Tap) error {
```
jcrussell: Remove and just call `t.destroyQos()`?

src/bridge/qos.go
```diff
@@ -0,0 +1,210 @@
...
+
+	t, ok := b.taps[tap]
+	if !ok {
+		return fmt.Errorf("tap %s not found", tap)
+	}
+
+	return b.updateQos(t, op)
+}
+
+func (b *Bridge) updateQos(t *Tap, op QosOption) error {
```
jcrussell: Same as `clearQos`.

src/minimega/qos_cli.go
```diff
@@ -0,0 +1,158 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Used to calulate burst rate for the token bucket filter qdisc
+const KERNEL_TIMER_FREQ uint64 = 250
+const MIN_BURST_SIZE uint64 = 2048
+
+var qosCLIHandlers = []minicli.Handler{
```
jcrussell: Wrap help to 80 columns.

src/minimega/vm.go
```diff
@@ -574,6 +677,13 @@ func (vm *BaseVM) info(key string) (string, error) {
 			s := fmt.Sprintf("%.1f/%.1f (rx/tx MB/s)", v.RxRate, v.TxRate)
 			vals = append(vals, s)
 		}
+	case "qos":
+		for _, v := range vm.Networks {
+			s := vm.QosString(v.Bridge, v.Tap)
+			if s != "" {
```
jcrussell: Probably need to append so that tap names line up with the qos values.

jcrussell:
LGTM.

@djfritz: want to test/merge this?

djfritz:
LGTM

PR #424, opened 2016-02-25
Add Is__() methods to patternItem

michaelwisely:
@jcrussell Is there any issue with exporting `itemType`? I have it redefined it in the Python API generator, but this seems like a better option.

Everything still builds for me, and tests are passing.

### DIFF
--- a/src/minicli/pattern.go
+++ b/src/minicli/pattern.go
@@ -46,11 +46,27 @@ type patternItem struct {
 
 type patternItems []patternItem
 
-func (p patternItem) Optional() bool {
+func (p patternItem) IsOptional() bool {
 	return p.Type&optionalItem != 0
 }
 
-func (p patternItem) List() bool {
+func (p patternItem) IsListeral() bool {
+	return p.Type&literalItem != 0
+}
+
+func (p patternItem) IsCommand() bool {
+	return p.Type&commandItem != 0
+}
+
+func (p patternItem) IsString() bool {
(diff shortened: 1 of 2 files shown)


djfritz:
LGTM unless @jcrussell objects

jcrussell:
I didn't export them because they're an implementation detail that's not necessary for users of minicli. I guess they're somewhat needed here... ugh, wwrpd?

michaelwisely:
Maybe it would make more sense to add a handful of methods to `patternItem` instead? I noticed [`Optional()` and `List()`](https://github.com/sandia-minimega/minimega/blob/master/src/minicli/pattern.go#L49) are there already. 

Either way is OK with me. In fact, the predicate methods might make more sense.

djfritz:
Rob wouldn't be afraid to redeclare some internal consts in another package
if it meant consistency and readability were maintained.

djfritz:
Looking back at this I think Mike's idea makes the most sense. Add
Command(),  String() (uh oh that's special), Choice(), and Literal() to the
others already there.

Because of the collision with the stringer interface maybe we should change
all of those to IsOptional...

michaelwisely:
@djfritz Can do. I'll revert this commit and squash with `IsThing()` methods.

djfritz:
LGTM

djfritz:
For good measure, there is precedence with this scheme. See golang `filemode.IsDir()` and `filemode.IsRegular()`

PR #425, opened 2016-02-25
minimega: add ability to skip over the vnc wait period

floren:
VNC playback files consist of one line per mouse or keyboard event. Each
line is associated with a delay period; minimega waits for the specified
delay, then sends the mouse or keyboard event to the VM. "vnc skip"
allows the user to immediately send the pending event without waiting
for the entire delay period.

### DIFF
--- a/src/minimega/vnc.go
+++ b/src/minimega/vnc.go
@@ -32,6 +32,7 @@ type vncClient struct {
 	Rhost string
 
 	done chan bool
+	skip chan bool
 	file *os.File
 
 	err error
@@ -78,6 +79,7 @@ func NewVNCClient(host, vm string) (*vncClient, error) {
 		Name:  vmName,
 		ID:    vmID,
 		done:  make(chan bool),
+		skip:  make(chan bool),
 	}
 
 	return c, nil
@@ -229,6 +231,7 @@ func (v *vncKBPlayback) playFile() {
 		wait := time.After(duration)
 		select {
 		case <-wait:
+		case <-v.skip:
 		case <-v.done:
 			return
 		}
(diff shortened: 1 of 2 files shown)


jcrussell:
When I saw this, I thought you meant a `fast-forward` capability -- skip ahead 100ms or something. As currently implemented, this seems like an odd functionality. /2c.

On a technical note, it looks like this could cause a deadlock if the playback finishes as the user is calling `vnc skip` since nothing will read off the skip channel.

djfritz:
The customer interest in record/replay is big and growing, so don't shoot it down just yet. 

I don't think this goes far enough however. Given the interest, I suggest we gut the current API and add a more robust one:

    vnc fb record <vm> <filename>
    vnc kb record <vm> <filename>
    vnc fb stop <vm>
    vnc kb stop <vm>

    # show current recordings/playbacks. More info is needed at least on playbacks (time remaining?, % complete...) as well as recordings (time recorded...)
    vnc     

    # start a playback or restart from the beginning of an existing playback
    vnc play <vm> <filename>
    
    # stop playback, but keep the current position in the playback file
    vnc stop <vm>
    
    # continue a stopped playback
    vnc continue <vm>
    
    # if started, skip the current delay and advance immediately to the next event (@floren's need)
    # if stopped, simply play the next event and stop again
    vnc step <vm>

I suggest we reject this PR, let @floren use it for the current need outside of mainline, and work towards something like this. @cdshann is a good candidate to implement after QoS and nuke patches get merged.

@floren comments?

floren:
Works for me, as Fritz says I implemented it for a specific need. The reasoning is that you frequently end up waiting longer than needed because you want to make sure whatever you're trying to click is actually there. However, if you're sitting there watching a playback, you can safely say "oh yeah it's ready, no need to wait another 2 minutes".

I like the "step" idea. It fulfills the same need but also adds a nice capability for a "stopped" playback.

djfritz:
My suggested patterns are slightly wrong, they need and host and vm name
instead of just a VM name (though we should have  convenience patterns for
localhost if you specify just a VM).

@floren want to register a new issue and close this?

PR #428, opened 2016-02-29
minimega: port fix from namespace branch.

### LINKED ISSUE #427: Deadlock in vm launch
Found this while working on the nuke fixes. To reproduce:

```
vm config net 100
tap create 100 ip 10.0.0.1/24
dnsmasq start 10.0.0.1 10.0.0.2 10.0.0.254

# this disk should not exist
vm config disk bad_disk.qcow2

vm launch kvm badvm # should return an error
vm launch kvm badvm # should return duplicate error

vm config disk valid_disk.qcow2
vm launch kvm goodvm # hangs indefinitely
```

jcrussell:
Fix deadlock when there's a duplicate VM name. Fixes #427.

### DIFF
--- a/src/minimega/vmlist.go
+++ b/src/minimega/vmlist.go
@@ -169,8 +169,18 @@ func (vms VMs) findVm(s string) VM {
 }
 
 // launch one VM of a given type.
-func (vms VMs) launch(vm VM) error {
+func (vms VMs) launch(vm VM) (err error) {
+	// Actually launch the VM from a defered func when there's no error. This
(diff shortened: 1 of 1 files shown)


cdshann:
Confirmed this fixes #427

djfritz:
The non-reporting of errors should have been fixed in https://github.com/sandia-minimega/minimega/commit/00c678f2027ebf4b8fa7dcb6028bbf337df6bcc2

If it's not then we need to fix that too. 

This same bug was possibly fixed in the container code in https://github.com/sandia-minimega/minimega/pull/373

PR #430, opened 2016-02-29
protonuke: fix http/https handlers and timing logs

floren:
Protonuke had a long-standing bug where requesting /12345/image.png
would return an HTML page rather than image.png as expected. This
was due to the way image handlers work in Go's http lib. This fixes
that behavior and also makes the client give more information when
logging GET requests.


djfritz:
LGTM

PR #432, opened 2016-03-02
Fixes issue that would fail to launch containers

x64x6a:
On latest debian, minimega would fail to mount minicgroup:
```
$ vm config filesystem /root/containerfs
$ vm launch container foo
2016/03/01 18:39:12 ERROR container.go:253: cgroup mount: device or resource busy
2016/03/01 18:39:12 ERROR container.go:721: cannot launch container VMs -- cgroups failed to initialize
2016/03/01 18:39:12 ERROR vm.go:487: write instance state file: open /tmp/minimega/0/state: no such file or directory
Error (debian): cannot launch container VMs -- cgroups failed to initialize
```

I was unable to mount as well:
```
# mount -t cgroup minicgroup foo/
mount: minicgroup is already mounted or /tmp/foo busy
```

I'm not too familiar with cgroups, but after some research and testing, I found that cgroup is mounted with tmpfs, but systemd is still mounted with cgroup.

I ended up using tmpfs instead of cgroup for the mount call in [container.go](https://github.com/sandia-minimega/minimega/blob/master/src/minimega/container.go), and minimega containers  would then load without any errors.

I'm not sure if this necessarily fixes the issue, but it seemed to be a possible solution. 

Any input regarding this?

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -267,7 +267,7 @@ func (_ ContainerIniter) init() error {
 		return fmt.Errorf("cgroup mkdir: %v", err)
 	}
 
-	err = syscall.Mount("minicgroup", CGROUP_ROOT, "cgroup", 0, "")
+	err = syscall.Mount("minicgroup", CGROUP_ROOT, "tmpfs", 0, "")
 	if err != nil {
 		return fmt.Errorf("cgroup mount: %v", err)
 	}


djfritz:
This patch simply disables cgroups for minimega. The cgroup vfs is a
special filesystem like procfs and sysfs. By mounting a tmpfs instead, all
minimega does is create some files in a directory.

The underlying problem is that systemd creates the root cgroup hierarchy
and minimega attempts to create one as well. That actually shouldn't be an
issue. Multiple root hierarchies are allowed, but systemd is doing
something (probably stupid).

We need to close this and open a ticket to look into systemd madness with
cgroups.

On latest debian, minimega would fail to mount minicgroup:

$ vm config filesystem /root/containerfs
$ vm launch container foo
2016/03/01 18:39:12 ERROR container.go:253: cgroup mount: device or
resource busy
2016/03/01 18:39:12 ERROR container.go:721: cannot launch container
VMs -- cgroups failed to initialize
2016/03/01 18:39:12 ERROR vm.go:487: write instance state file: open
/tmp/minimega/0/state: no such file or directory
Error (debian): cannot launch container VMs -- cgroups failed to initialize

I was unable to mount as well:

# mount -t cgroup minicgroup foo/
mount: minicgroup is already mounted or /tmp/foo busy

I'm not too familiar with cgroups, but after some research and testing, I
found that cgroup is mounted with tmpfs, but systemd is still mounted with
cgroup.

I ended up using tmpfs instead of cgroup for the mount call in container.go
<https://github.com/sandia-minimega/minimega/blob/master/src/minimega/container.go>,
and minimega containers would then load without any errors.

I'm not sure if this necessarily fixes the issue, but it seemed to be a
possible solution.

Any input regarding this?
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/sandia-minimega/minimega/pull/432
Commit Summary

   - Changed fstype to tmpfs from cgroup

File Changes

   - *M* src/minimega/container.go
   <https://github.com/sandia-minimega/minimega/pull/432/files#diff-0> (2)

Patch Links:

   - https://github.com/sandia-minimega/minimega/pull/432.patch
   - https://github.com/sandia-minimega/minimega/pull/432.diff

—
Reply to this email directly or view it on GitHub
<https://github.com/sandia-minimega/minimega/pull/432>.

x64x6a:
Gotcha.  That makes more sense.

PR #434, opened 2016-03-02
minimega: adding `vm config tag`.

### LINKED ISSUE #429: minimega: add `vm config tag`
`vm config tag` should work just like `vm tag` except that it applies to launched VMs.

jcrussell:
Allows you to configure tags for all newly launched VMs. Fixes #429.


djfritz:
Man that's a big change to support this. I see that nearly all of the change is to support clearing specific tags. Do you see future use cases for Clear() needing a handle to the calling command in other APIs (not that I see a clearer way to do this, just asking)?

jcrussell:
Yeah, it ended up being a bit bigger than I had hoped. I could create a separate `clear vm tag [key]` handler so that we don't have to change the function definition for clear. We could use this in the following patterns:

```
clear vm config qemu-override [id]
clear vm config disk [path to disk image]...
```

The first would replace the `vm config qemu-override delete <id or all>` command which is inconsistent with the rest of the API. The second would remove one or more disk images from a VM.

djfritz:
I like that solution. Want to take a crack at it?

jcrussell:
To clarify, you want the two new patterns and not the separate clear handler, correct?

If so, I should be able to do that today.

djfritz:
Yea I think that's cleaner. Also, other APIs that don't use configfns helpers are already going to have to use the approach you suggested, so we should just do it everywhere.

jcrussell:
Hmm... thinking about this again. What if we split this off into a separate command:

```
qemu # prints qemu launch string based on current VMConfig
qemu override # tabular output
qemu override <match> <replacement> # add a new override
clear qemu override [id] # could replace this with match
```

I'm wondering if this makes more sense than embedding it in the `vm config` API. There's a few places in the tests where I call `vm config qemu-override` just to verify that the string looks the same as it did before (particularly useful for 40+ virtio devices). At the very least it would be useful to have a way to print the qemu string without all the extra output from `vm config qemu-override`.

Thoughts?

jcrussell:
@djfritz: think this is ready for review now.

djfritz:
A few things:

- the modified clear handler is still there
- what is `debug qemu-args` for? The debug API is for debugging information for minimega, having a vm config thing in there is confusing. Leaving this in the qemu-overrides is fine.
- We shouldn't sneak in things like entirely new APIs in PRs fixing unrelated things.

jcrussell:
* Didn't we agree on keeping the modified clear handler?
* It seems cleaner to have a separate command to print out the qemu string with and without overrides. It cleaned up the tests and simplified the code so it seemed useful. 
* The scope grew and I should have renamed the PR.

PR #440, opened 2016-03-16
minimega: adding `vm config tag`.

jcrussell:
Allows you to configure tags for all newly launched VMs. Fixes #429.

PR #441, opened 2016-03-16
minitweaks to minimega/minitest.

jcrussell:
_Merge after #440._

Change default print behavior for vmConfigSlice to print an empty string when the slice is empty.

Changed minitest to print an empty line when it reads an empty line.

Updated all tests accordingly.


djfritz:
LGTM

PR #443, opened 2016-03-17
minimega: fixes for disk API.

jcrussell:
Relative paths should always be relative to /files/, not a mixture of /files/ and CWD. Adding tests to make sure we can create/snapshot/inject images.


djfritz:
LGTM

PR #445, opened 2016-03-18
minimega/tests: fix tap tests.

jcrussell:
Changed defunctTaps from a slice to a map so that we can omit printing
of defunctTaps. This removes a race if we call `tap delete all` and then
immediately call `tap`.

Updated simple lifecycle test for taps.


djfritz:
LGTM

PR #446, opened 2016-03-18
minimega/minitest/tests: fix `vm save`.

jcrussell:
Lots of changes to get `vm save` working properly. Specifically:

 * remove `kvm` and `container` annotations from `vm config` commands --
   we experimented with `vm <vm type> config ...` but ended up simply
   omitting the VM type. Updated documentation where needed.
 * fix saved launch commands (we now have to supply a VM type)
 * add `vm config cpu` from namespace branch -- it was almost
   configurable in master which caused it to show up in the output of
   `vm save`. Rather than add extra code to work around this, just added
   the new API.
 * adding container-specific configurations to output of `vm config` so
   that we can inspect them.
 * fix bug in minitest where we printed the original command too often.
 * add tests for `vm save`.


djfritz:
LGTM

PR #453, opened 2016-03-21
web: fix hide connect button on non-kvm VMs.

jcrussell:
Set style rather than node attr. Tested on Chrome/Firefox.


djfritz:
LGTM

PR #456, opened 2016-03-21
minimega/miniclient: move Attach to minimega.

### LINKED ISSUE #455: minidoc: revert to custom minicli interface
https://github.com/sandia-minimega/minimega/pull/376 broke the appengine build for minidoc, as it imports goreadline, os, os/signal, syscall, and others that are not available in appengine.

This was fixed once before with https://github.com/sandia-minimega/minimega/pull/286 . I let this slip by untested.

@jcrussell I'm going to cut this in as in #286 unless you have a quick idea.

jcrussell:
Too many dependencies for app engine. Moving code to minimega so that we
can still import miniclient in app engine for minidoc.

Should fix #455.

### DIFF
--- a/src/miniclient/client.go
+++ b/src/miniclient/client.go
@@ -7,16 +7,13 @@ package miniclient
 import (
 	"encoding/json"
 	"fmt"
-	"goreadline"
 	"io"
 	"minicli"
 	log "minilog"
 	"minipager"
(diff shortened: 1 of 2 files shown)


djfritz:
miniclient imports minipager which imports syscall. No go.

PR #463, opened 2016-03-24
minidoc: only display non-empty dirs.

jcrussell:
minidoc now checks to make sure there is at least one shown file in a
directory before listing it as a sub-directory.

### DIFF
--- a/doc/content/articles/namespaces.article
+++ b/doc/content/articles/namespaces.article
@@ -1,0 +1,190 @@
+Namespaces
+
+The minimega authors
+22 Mar 2016
+
+* Introduction
+
+One of the early feature requests for minimega was a scheduler that would launch
+VMs across a cluster of machines as easily as VMs are launched on a single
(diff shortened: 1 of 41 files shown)


djfritz:
If we use this, we'll have to maintain the patch as this is vendorized code from golang's present tool. We already have to patch with minimega.go in order to register the .mega directive. 

Knowing this, do we still want to proceed?

djfritz:
Alternatively, we could just flatten the directory structure of our docs. We don't need to put examples/images/etc. in subdirectories.

jcrussell:
It looks like we tweak minidoc a few times a year at most and we haven't pulled in the latest code from the present tool in a long time (although present itself has only added about 100 lines since July 2014).

It doesn't seem like much effort to maintain this patch but if we'd rather reorganize, that works too. If it did become a burden to maintain the patch, we could always reorganize at that point.

PR #466, opened 2016-03-24
minidoc: only display non-empty dirs

jcrussell:
minidoc now checks to make sure there is at least one shown file in a
directory before listing it as a sub-directory.

Accidentally killed the previous commit for this (#463).


djfritz:
LGTM

PR #471, opened 2016-03-24
minimega: fix #467.

### LINKED ISSUE #467: minimega: crash with namespaces
This crashes minimega:

```
namespace foo nsmod add-host localhost
namespace foo vm launch kvm 1
namespace foo vm launch
```

Have a patch ready.

jcrussell:
The following commands would crash minimega:

    namespace foo nsmod add-host localhost
    namespace foo vm launch kvm 1
    namespace foo vm launch

Because the active namespace was reverted before `vm`launch` was
completed (asynchronously). Copied the active namespace to the local
scope to fix.


djfritz:
LGTM?

Issue #472, opened 2016-03-25
minimega: consider lock refactoring


djfritz:
Here's the current list of all locks in `src/minimega`, with some notes/thoughts. This is beginning to get away from us and we should consider paths forward before adding more layers of complications.

Locks that are okay/necessary: 

- capture.go - global mutex that wraps the global captureEntries map
- misc.go - loggingMutex type for debugging (Jon is insane)
- container.go - global mutex wrapping containerInit()
- vlans.go - mutex in AllocatedVLANs type that wraps map accesses
- command_meshage.go - mutex that serializes meshageSend() calls
- stats.go - global mutex that wraps bandwidthStats map
- namespace.go - waitgroup in namespaceLaunch(), used to wait for VM scheduling to finish

Locks that need some toughts/comments:

- vmlist.go - waitgroup and mutex in VMs.apply() - the apply() fanout functions seem unnecessarily complex to me in the first place. It's very compact, but I wonder how well we can maintain this moving forward vs some readable pasta. 
- vm.go - per VM mutex in BaseVM type - used all over the place. Necessary but the usage should be simplified/codified.
- kvm.go - sync.Once used in isNetworkDriver() - Not hurting anything but the use of sync.Once and function literals seems so unnecessarily complex to me.
- vm_cli.go - waitgroup used to block on acks for `vm launch`. I think we need to rethink the concept of VMs ack'ing on launch. 
- vm.go - global vmLaunch waitgroup - synchronizes noblock vms and used in cli.go. Should maybe move waitgroup calls into launch routines or rethink exactly how we do launching in order to remove the need for the noblock suffix.
- vm.go - global vmLock mutex - used in preamble routines (checkInterfaces(), in vmlist.go, container.go, kvm.go...). Necessary, but usage should be simplified, call path should be flattened maybe so simplify who should be grabbing the lock.
- bridge.go - *five* locks, three global, two in the Bridge type. This is nuts. Just serialize all bridge *everything* with a global bridge lock and move on. There is just no way that this is in some crazy critical path that needs all of this.

PR #473, opened 2016-03-25
minimega: fix #469.

### LINKED ISSUE #469: minimega: vm start should wait for vm launch
Don't think this works anymore with namespaces.

jcrussell:
Much saner locking so that we can:

 * block on `vm start all` while `vm launch ... noblock` launches things
   in the background
 * block on `vm launch` while the scheduler runs

Removes vmLaunch lock and instead pre-locks the VMs on creation. This
ensures that Launch() is the first operation called on a VM (otherwise
there will be a deadlock).

Updates #472.


djfritz:
LGTM

Issue #478, opened 2016-03-30
minimega: OOM killed


jcrussell:
Hammered on minimega for about 10 hours and it was OOM killed. I suspect that it may be the command history (`minicli.history`) which grows indefinitely that caused it to run OOM. We should probably limit the history and then see if we can reproduce this.

djfritz:
HA!

jcrussell:
Thinking about an upgrade to the `debug` API:

```
            "debug",               // outputs the stats from runtime
            "debug memory <file>", // dump memory profile to file
```

This would make it easy to dump the memory profile every hour until it crashes. Thoughts?

djfritz:
Sounds good to me. We could add a CPU profile too while we're at it.

jcrussell:
CPU profiles are stopped/started -- we could have the length of the snapshot as a parameter to simplify the API (no stop/start commands).

```
            "debug <cpu,> <duration> <file>", // dump CPU profile for duration seconds to file
```

djfritz:
Up to you, we can defer adding CPU profiles if need be. Or we can enable it from the debug API and have the stop occur in the teardown() signal handler

djfritz:
Is this fixed now? @jcrussell

djfritz:
Oh wait, no. I wasn't paying attention.

jcrussell:
Merging #481 and #484 will close this.

jcrussell:
Both tickets merged... Closing ticket.

PR #479, opened 2016-03-30
minimega: add profiling to `debug` API

jcrussell:
Add `debug memory` and `debug cpu` APIs to dump profiling information
about minimega.

Updates #478


djfritz:
LGTM

PR #480, opened 2016-04-01
protonuke: timing & TLS

floren:
Undo some timing changes I'd made; we want to only count http/https "pageloads" rather than individual GETs.

Also added capability to tell the client what TLS version to use.

Sorry these are in the same PR

### DIFF
--- a/src/protonuke/http.go
+++ b/src/protonuke/http.go
@@ -59,7 +59,7 @@ func httpClient(protocol string) {
 		Proxy: http.ProxyFromEnvironment,
 		Dial: func(network, addr string) (net.Conn, error) {
 			dialer := &net.Dialer{
-				Timeout: 30 * time.Second,
+				Timeout:   30 * time.Second,
 				KeepAlive: 30 * time.Second,
 			}
 			return dialer.Dial(protocol, addr)
@@ -69,7 +69,7 @@ func httpClient(protocol string) {
 	// TODO: max client read timeouts configurable?
 	client := &http.Client{
 		Transport: transport,
-		Timeout: 30 * time.Second,
+		Timeout:   30 * time.Second,
 	}
 
 	for {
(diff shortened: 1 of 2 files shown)


floren:
@djfritz

src/protonuke/protonuke.go
```diff
@@ -37,6 +37,7 @@ var (
 	f_httpImageSize = flag.Int("httpimagesize", 3, "size of image, in megabytes, to serve in http/https pages")
 	f_httpTLSCert   = flag.String("httptlscert", "", "file containing public certificate for TLS")
 	f_httpTLSKey    = flag.String("httptlskey", "", "file containing private key for TLS")
+	f_tlsVersion		= flag.String("tlsversion", "", "Select a TLS version for the client: tls1.0, tls1.1, tls1.2")
```
jcrussell: `go fmt`?

src/protonuke/http.go
```diff
@@ -99,6 +99,20 @@ func httpTLSClient(protocol string) {
...
+			version = tls.VersionTLS10
+		case "tls1.1":
+			version = tls.VersionTLS11
+		case "tls1.2":
+			version = tls.VersionTLS12
+		}
+		transport.TLSClientConfig.MinVersion = tls.VersionSSL30
+		transport.TLSClientConfig.MaxVersion = version
+	}
+
```
jcrussell: Are max and min reversed?

djfritz: Shouldn't be. TLS 1.0 > SSL 3.0 in that history.

src/protonuke/protonuke.go
djfritz: Update help text and bracket options:

   Select a maximum client TLS version [1.0, 1.1, 1.2]

djfritz:
Can you update the PR to not include the protonuke binary?

djfritz:
LGTM

PR #484, opened 2016-04-06
minicli: limit history size

jcrussell:
Truncate the history stored in minicli to a maximum size. Developers can
tweak the maximum as needed or disable limit entirely, if needed.

### DIFF
--- a/src/minicli/minicli.go
+++ b/src/minicli/minicli.go
@@ -46,6 +46,15 @@ var defaultFlags = Flags{
 var handlers []*Handler
 var history []string // command history for the write command
 
+// HistoryLen is the length of the history of commands that minicli stores.
+// This may be increased or decreased as needed. If set to 0 or less, the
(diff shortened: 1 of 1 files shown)


src/minicli/minicli.go
```diff
@@ -46,6 +46,15 @@ var defaultFlags = Flags{
 var handlers []*Handler
 var history []string // command history for the write command
 
+// HistoryLen is the length of the history of commands that minicli stores.
+// This may be increased or decreased as needed. If set to 0 or less, the
+// history will grow unbounded and may cause an OOM crash.
+var HistoryLen = 10000
```
djfritz: should this be a constant instead?

jcrussell:
I thought about const but then I figured at some point we would have to add an API to minimega to set the history length based on the need to store HistoryLen+1 commands in the history.

djfritz:
+1 for appropriate level of cynicism. LGTM

PR #485, opened 2016-04-06
minimega: simplify isMac.

jcrussell:
Use net.ParseMAC rather than a regular expression. Cleaner and faster:

BenchmarkIsMac           50000             30745 ns/op
BenchmarkIsMac2        3000000               521 ns/op
BenchmarkIsMacNet     10000000               160 ns/op

Second benchmark uses a pre-compiled regular expression.


djfritz:
That's a pretty serious speedup! LGTM

PR #490, opened 2016-04-13
ron/minimega: cc++

### LINKED ISSUE #488: ron/miniccc: add negation to `cc filter`
Before namespaces, we could launch all the clients on a single host and run `cc background` without any filters to make them all start `protonuke`. Now, targeting these VMs is more complicated because the clients could be split between multiple hosts. We can currently specify a CIDR for ip (e.g. 100.0.1.0/24). It would be nice to also be able to specify ips to excluded, such as the router on 100.0.1.1. The syntax would be something like `ip=100.0.1.0/24 ip!=100.0.1.1`.

jcrussell:
Summary:

 * Add `cc filter tag=`
 * Fix crash caused by implicit filtering
 * Fix race with VM tags
 * Add better filtering for namespace-specific cc commands/responses.

Closes #488.

### DIFF
--- a/src/minimega/cc.go
+++ b/src/minimega/cc.go
@@ -105,6 +105,18 @@ func ccClients() map[string]bool {
 	return nil
 }
 
+// ccGetFilter returns a filter for cc clients, adding the implicit namespace
+// filter, if a namespace is active.
+func ccGetFilter() *ron.Client {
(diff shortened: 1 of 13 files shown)


djfritz:
LGTM!

PR #491, opened 2016-04-13
minimega: add `check` flag to `read`

jcrussell:
`validate` reads a mm script and validates the syntax of the commands
therein. Only checks syntax and not semantics.


djfritz:
LGTM, but was is this used for?

jcrussell:
I generated a `launch.mm` script using a `launch.bash` script and I wanted to check that all the commands would compile before I ran it. We could merge this into the `read` API and have an optional flag to validate but not run the commands.

jcrussell:
Updated to add a flag to `read`. Much fewer LoC for the same functionality.

djfritz:
LGTM

PR #494, opened 2016-04-14
minimega: fix races and refactoring

jcrussell:
Lots of refactoring to fix several known races and a few potential races. Most of the work went into ensuring that all uses of `vms` are synchronized using `vmLock`. Moved all uses of `vmLock` to `vmlist.go` and followed naming convention for all `VMs` methods:

 * capitalized (would-be exported) functions acquire `vmLock`
 * lower-cased functions assume `vmLock` is held by caller

Also used `go install -race minimega` to identify two other data races which are now fixed. This passes all the `minitest` cases -- I think it should be at least as stable as master.


jcrussell:
Passes all the new minitests.

jcrussell:
Don't merge: this hung when I tried launching an enterprise-level experiment.

jcrussell:
This is finally ready to merge.

@djfritz

djfritz:
Working on this today.

djfritz:
Cursory review LGTM. I'll do some random tests on it, but this looks like it's well captured in minitest.

djfritz:
LGTM

PR #498, opened 2016-04-19
minifuzzer: new tool to fuzz minimega.

jcrussell:
Uses the `-cli` flag to figure out the list of commands that it can run.
Randomly select patterns and then generates the command string to match.
Little scared to test it on my workstation...


jcrussell:
We should probably test this on CCC before merging.

djfritz:
LGTM, @jcrussell can I merge this?

Issue #468, opened 2016-03-24
minimega: tap/bridges in namespaces


jcrussell:
We should figure out how taps and bridges interact with namespaces. Should they be handled like VLANs (which automatically prefix the alias with the active namespace)? We will have to figure out how this works with multiple hosts too.

jcrussell:
This is difficult since bridges have a physical backing.. will think about later if needed.

PR #501, opened 2016-04-19
minimega/bridge: huge refactoring

jcrussell:
_Note: merge after #423, #500_

Huge refactoring to split the bridge code out into a separate package to help make the tap/bridge APIs namespace aware (#468). This also allowed for heavy refactoring of how the locking in the bridge code works to make it much saner (#472).

This passes all the minitest cases that we have currently.


jcrussell:
TODO: move bandwidth stats to bridge package.

jcrussell:
@djfritz: I think this is ready to merge.

djfritz:
There's no good way to review a PR this big. I looked over it a week or so ago and it seemed sane... LGTM and let's find out.

PR #500, opened 2016-04-19
minimega: fix nits

jcrussell:
_Note: merge after #491, #494._

I've been wanting to change the argument type for the `wrap(Simple|Broadcast|VMTarget)CLI` functions to simplify writing handlers. As you can see from the LoC changes, this removes a lot of boilerplate code.

I also (finally) figured out an approach to resolve the command duality in namespaces. See commit message and documentation in `wrapBroadcastCLI` for an explanation.

This passes all the minitest cases that we have currently.

PR #502, opened 2016-04-20
tests/minimega: few new tests and minor tweaks to minimega

jcrussell:
Added a few tests for the following APIs:

 * `check`
 * `debug`
 * `file`
 * `shell`
 * `vm cdrom`
 * `vm hotplug`
 * `vm qmp`
 * `vm migrate`
 * `vm screenshot`

Made a couple of minor tweaks to minimega:

 * Change `vm migrate` column names to be consistent with `vm info`
 * Don't record `clear history` in history

This should be merged *before* #494 and #500 so that we can apply these tests to that code.


djfritz:
LGTM

Issue #470, opened 2016-03-24
doc: write developers guide to namespaces


jcrussell:
Explain why and how everything works.

PR #503, opened 2016-04-22
doc: add namespace article for developers.

jcrussell:
Initial cut of developer docs for namespaces. Updates #470.

@djfritz: this is everything I can think of so far, any thoughts on completeness? Maybe will have some more ideas next week.

This ticket assumes that all the in-flight PRs get merged (although it's not too far off the `master` implementation).


djfritz:
LGTM

PR #505, opened 2016-04-25
minimega: remove meshage timeout.

### LINKED ISSUE #504: minimega/mesh: `vm start` timeout
When you call `vm start all`, it blocks until all the VMs that were launched with `vm launch ... noblock` have been created and started. This may cause timeouts with `mesh send`.

I propose that we get rid of this timeout -- we should trust that the node responds or crashes. If the node crashes, we should see an error from the connection's reader.

Perhaps we could support interrupting "long-running" commands with `Ctrl-C` if commands truly get stuck (e.g. `mesh send` and `read`).

jcrussell:
Fixes #504. Set the default timeout to 0 which implies no timeout. Users
can set a timeout using `mesh timeout`, if needed.

### DIFF
--- a/src/minimega/meshage.go
+++ b/src/minimega/meshage.go
@@ -8,6 +8,7 @@ import (
 	"encoding/gob"
 	"fmt"
 	"iomeshage"
+	"math"
 	"math/rand"
 	"meshage"
 	"minicli"
@@ -16,10 +17,6 @@ import (
 	"reflect"
 	"time"
 	"version"
-)
-
-const (
-	MESH_TIMEOUT_DEFAULT = 10
 )
 
 type meshageCommand struct {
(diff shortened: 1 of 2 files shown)


src/minimega/meshage.go
```diff
@@ -168,8 +163,15 @@ func meshageSend(c *minicli.Command, hosts string) (chan minicli.Responses, erro
 
 		log.Debug("meshage sent, waiting on %d responses", len(recipients))
 
+		// host -> response
 		resps := map[string]*minicli.Response{}
 
+		timeout := meshageTimeout
+		// If the timeout is 0, set to "unlimited"
+		if timeout == 0 {
+			timeout = math.MaxInt64
```
djfritz: But what if I issue a command that naturally takes longer than the expected remaining lifetime of the sun?

djfritz:
LGTM

src/minimega/meshage.go
jcrussell: Hah. Fortunately, time.Duration is a nanosecond count so that's only ~290 years.

PR #508, opened 2016-04-25
minimega: container tweaks

### LINKED ISSUE #507: minimega: launching container without FS should be an error
We don't check that the FS is set for containers but the `vm launch` will always fail when we try to create the overlay. We should enforce this earlier so that we can return a less cryptic error than:

```
2016/04/25 20:00:38 ERROR container.go:1139: overlay mount: exit status 32 mount: wrong fs type, bad option, bad superblock on megamount_3,
       missing codepage or helper program, or other error

       In some cases useful info is found in syslog - try
       dmesg | tail or so.

2016/04/25 20:00:38 ERROR container.go:770: overlayMount: exit status 32
2016/04/25 20:00:38 DEBUG vm.go:555: updating vm 3 state: BUILDING -> ERROR
```

jcrussell:
Fixes #507. Return an error if the filesystem is not set for containers.

Tweak `vm config init` so that it uses the `vmConfigSlice` helper (specifically, needed `PrintCLI`). Fixes an issue when trying to launch containers with namespaces.

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


djfritz:
LGTM

PR #509, opened 2016-04-25
minicli: improved .filter builtin

jcrussell:
Now includes four patterns for filters:

```
.filter host=ccc1 vm info
.filter host!=ccc1 vm info
.filter host~ccc vm info
.filter host!~ccc vm info
```

The latter two being substring matching. Added implicit substring matching for fields that appear to be JSON objects (e.g. `tags` in `vm info`) -- we already had this for fields that appear to be lists. Also made the code a lot more readable.

### DIFF
--- a/src/minicli/builtins.go
+++ b/src/minicli/builtins.go
@@ -5,11 +5,18 @@
 package minicli
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


djfritz:
What no tests?

jcrussell:
@djfritz: an oversight on my part, for sure.

PR #521, opened 2016-05-10
iomeshage: fix #519

### LINKED ISSUE #519: iomeshage: fix deadlock
#518 fixed the concurrent map access but introduced a deadlock. Need to release the TIDLock before sending on the channel.

jcrussell:
Release TIDLock before sending on channel. Move TIDLock into IOMeshage
struct and rename to tidLock so that it is no longer exported.


jcrussell:
Tested on CCC, seems good.

djfritz:
LGTM

PR #522, opened 2016-05-10
minimega: fix #520

### LINKED ISSUE #520: Update help api
```
 For example, to capture netflow data on bridge mega_bridge to file in ascii mode and with gzip compression:

capture netflow mega_bridge file foo.netflow ascii gzip

You can change the active flow timeout with:

capture netflow mega_bridge timeout <timeout>
```

From the capture api example you need to use the word bridge before specifying a bridge
capture netflow bridge mega_bridge file foo.netflow ascii gzip

jcrussell:
Update help text for capture API.


djfritz:
LGTM

PR #523, opened 2016-05-10
minimega: add new preprocessor

### LINKED ISSUE #514: minimega: add `http://` preprocessor
Add preprocessor to download files via HTTP. For example:

```
vm config disk http://minimega.org/examples/client.qcow2
```

The file should be downloaded to the `files` directory. We should check if a file with the same name already exists in the `files` directory (or in another node's `files` directory) and use that instead of redownloading the file.

jcrussell:
Add preprocessor for http:// and https://. Gutted the file:
preprocessor. Closes #514.


djfritz:
We should update the file.article to reflect this preprocessor as well.

jcrussell:
@djfritz: updated file.article.

djfritz:
LGTM

PR #529, opened 2016-05-13
minicli: add .alias and .unalias builtins

### LINKED ISSUE #528: minicli: `.alias` and `.unalias` API
Functions similar to bash aliases -- replace text at the beginning of the commands string using stored aliases. For example:

```
.alias vmr=.filter state=running vm info
vmr
```

We should `log.Info` on expansion.

jcrussell:
Works similar to bash aliases. Updated minitests. Fixes #528.


djfritz:
Of course you did this already...

djfritz:
LGTM

PR #530, opened 2016-05-13
minimega: add `vm summary`

### LINKED ISSUE #510: minimega: `vm info` columns
`vm info` is getting unruly -- thankfully, we have `.filter` and `.columns`. Some useful fields are missing (e.g. `filesystem` and `hostname` for containers) and some fields can be huge (e.g. `tags`). We certainly need to add the container-specific columns but we already have 24 columns -- when should we stop?

Thoughts?

@djfritz @floren

jcrussell:
Fewer columns than `vm info` but with the same meanings. Fixes #510.

@floren @djfritz: these seemed like a reasonable subset of the columns, thoughts?

### DIFF
--- a/src/minimega/vm.go
+++ b/src/minimega/vm.go
@@ -130,12 +130,17 @@ type BaseVM struct {
 	instancePath string
 }
 
-// Valid names for output masks for vm info, in preferred output order
-var vmMasks = []string{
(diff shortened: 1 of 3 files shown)


djfritz:
LGTM @floren ?

djfritz:
Well this still LGTM, merging.

PR #532, opened 2016-05-13
minimega: add `cc process kill all`

### LINKED ISSUE #489: minimega: add 'kill all' to cc process api
see summary

jcrussell:
Add wildcard support to `cc process kill`. Fixes #489.

Going to test this today.

### DIFF
--- a/src/minimega/cc_cli.go
+++ b/src/minimega/cc_cli.go
@@ -437,16 +437,7 @@ func ccProcessKill(pid int) {
 func cliCCProcessKill(c *minicli.Command, resp *minicli.Response) error {
 	// kill all processes
 	if c.StringArgs["pid"] == Wildcard {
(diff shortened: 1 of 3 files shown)


djfritz:
This solution is a bit blunt - it spams out running PIDs to every client that matches on the filter which causes a lot of errors to come back which we don't care about. I'd rather add a magic byte (-1) to srn/ron for a PID wildcard and support 'kill all' by adding a case to:

https://github.com/sandia-minimega/minimega/blob/master/src/ron/client.go#L505

for checking for the 'all pids' wildcard. At least then 'kill all' becomes a single cc command and doesn't prompt unnecessary errors.

jcrussell:
@djfritz: added less blunt fix, will test today.

djfritz:
LGTM

PR #535, opened 2016-05-17
Recovery mode for web interface

mkunz7:
Parses the running qemu processes and creates a limited web interface.
Useful for when minimega crashes and you have put a lot of effort into
an experiment. Python.
 
capture12
vnc
autotile


djfritz:
What's causing minimega to crash that prompts this PR? I haven't seen a crash report in some time now.

mkunz7:
Didn't have enough logging enabled to have any idea. Mini crashes, VMs stay up. Prior it was the capture api killing things, not sure what caused it this time around. Maybe we should fund some interns to look at mini and report crashes.

jcrussell:
Stack traces are dumped to stderr, regardless of log level. If you redirect stderr to a file, we'll be able to figure out what caused the crash. If you typically run minimega with debug/info level logging and don't want to fill up the disk, you can disable logging to stderr with `log stderr false`.

That being said, I know at least one crash (#540) that I'm working on patching.

djfritz:
There's no way to use this in mainline. We're gutting the entire web interface on October 1, so this would be moot anyways.

PR #542, opened 2016-05-25
minidoc: change -port to -server

michaelwisely:
Now you can specify which host to listen on, if you're so inclined.


djfritz:
LGTM

jcrussell:
This breaks: https://github.com/sandia-minimega/minimega/blob/master/misc/daemon/minimega.init

PR #543, opened 2016-05-26
minimega: add vms column to hosts API

### LINKED ISSUE #539: minimega: add `vms` column to `host`?
Should we add a count of the number of vms to `host`? I launched a bunch of VMs with namespaces and it's non-trivial to figure out how many landed on each host. I know we /can/ get the info elsewhere but this seems useful to me.

Alternatively, we could add a `.count` API...

jcrussell:
Report the number of vms in the active namespace and across namespaces
for the host. Fixes #539.


djfritz:
LGTM

PR #544, opened 2016-05-26
minimega: fix VM nits

jcrussell:
Fix a bunch of smaller things that I didn't like about the VM interface.

Questions:
 * Do we want to warn if two VMs share the same MAC or return an error?
 * Can VMs share the same name if one is in the global namespace?
 * Should `vm save` record the namespace of the VM? (I think we want this)

@djfritz: thoughts?

### DIFF
--- a/src/minimega/capture.go
+++ b/src/minimega/capture.go
@@ -102,7 +102,7 @@ func startCapturePcap(vm string, iface int, filename string) error {
 		return vmNotFound(vm)
 	}
 
-	config := v.Config()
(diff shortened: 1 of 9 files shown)


jcrussell:
Phone discussion yielded:

* When is warn?
* What is allow?
* How is should?

jcrussell:
Passes all minitests. Going to try booting an enterprise once the cluster is free.

jcrussell:
Tested a couple of times at enterprise-scale -- seems to work.

PR #547, opened 2016-05-27
minimega: fix deadlock in namespace

### LINKED ISSUE #546: minimega: fix deadlock
`tests/taps_namespace` causes deadlock...

jcrussell:
Can't use hostTapDelete in Namespace.Destroy since it tries to query and
update the active namespace. Instead, call the methods to delete the tap
manually. Fixes #546.


djfritz:
LGTM

PR #548, opened 2016-05-27
minimega/bridge: fix capture

### LINKED ISSUE #545: Pcap capture 0 bytes
minimega  2016-05-08 works
minimega  2016-05-26 fails

Did a fresh install of latest same issue, something has changed.....

jcrussell:
Created a tuntap rather than letting ovs create the interface for us.
Capture now seems to work. Fixes #545.

### DIFF
--- a/src/bridge/mirror.go
+++ b/src/bridge/mirror.go
@@ -22,7 +22,7 @@ func (b *Bridge) CreateMirror() (string, error) {
 	}
 
 	// get a host tap on VLAN 0
-	tap, err := b.createTap("", 0, true)
+	tap, err := b.createHostTap("", 0)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


djfritz:
Why does this fix the issue? Was this just a race between invoking ovs creating a host tap and handing it to capture api for things?

src/bridge/mirror.go
```diff
@@ -22,7 +22,7 @@ func (b *Bridge) CreateMirror() (string, error) {
 	}
 
 	// get a host tap on VLAN 0
-	tap, err := b.createTap("", 0, true)
+	tap, err := b.createHostTap("", 0)
```
djfritz: Isn't this still having OVS create the tap via addTap()?

jcrussell:
The OVS `add-port` command generated by minimega for host taps includes `set Interface <name> type=internal`. Currently, that command will automatically create the tap for us (and remove it when we `del-port`). When I created the tuntap first using `ip` and then tried to add it via `add-port`, the tuntap won't see any traffic.

There's probably something going on with OVS that I don't fully understand, perhaps there's an easier workaround? I'm copying the behavior that we had before the bridge refactoring.

djfritz:
Woop, I saw this a "create a tuntap..." instead of "created a tuntap..." this makes sense now. LGTM once the merge conflict is fixed.

djfritz:
LGTM

PR #553, opened 2016-06-01
bridge/minimega: fix #540

### LINKED ISSUE #540: minimega: fix panic
```
panic: close of nil channel

goroutine 35577 [running]:
panic(0xac0460, 0xc82a9be9c0)
        /opt/go/src/runtime/panic.go:464 +0x3e6
ipmac.(*IPMacLearner).DelMac(0xc82ab5c700, 0xc82010d860, 0x11)
        /minimega/src/ipmac/ipmac.go:71 +0xbf
main.(*ContainerVM).launch.func1(0xc822a9c9a0, 0xc822849180, 0xc823ce56c0, 0x17)
        /minimega/src/minimega/container.go:1009 +0xb20
created by main.(*ContainerVM).launch
        /minimega/src/minimega/container.go:1028 +0x2cf2
```

jcrussell:
Created a local err in `Container.launch` when calling `bridges.Get`. If
this failed, we wouldn't register a MAC with the bridge so when we tried
to unregister it, minimega panicked. Move networking code to a separate
function so it's easier to return an error if there's a problem rather
than break out of the loop. Fixed some potential issues in the bridge
code when creating and destroying a bridge.

### DIFF
--- a/src/bridge/bridges.go
+++ b/src/bridge/bridges.go
@@ -83,21 +83,27 @@ func (b Bridges) newBridge(name string) error {
 
 	br.preExist = !created
 
-	// Bring the interface up
-	if err := upInterface(br.Name, false); err != nil {
+	// Bring the interface up, start MAC <-> IP learner
+	err = upInterface(br.Name, false)
+	if err == nil {
+		err = br.startIML()
+	}
+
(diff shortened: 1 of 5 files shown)


djfritz:
LGTM

PR #554, opened 2016-06-01
minimega: fix VM nits

jcrussell:
Fix a bunch of smaller things that I didn't like about the VM interface.

Need to test after #553 gets merged.


djfritz:
The new conflicts approach is way better than how we did this before. LGTM

PR #561, opened 2016-06-07
minimega: abort after error in read

### LINKED ISSUE #560: minimega: read should stop after error
Doesn't look like read stops after a command returns an error:

```
minimega$ shell cat test
foo: echo before
shell asdfasdfas
echo after
minimega$ read test
foo: before
Error (foo): exec: "asdfasdfas": executable file not found in $PATH
foo: after
```

jcrussell:
Actually stop reading the file if there's an error. Fixes #560.

### DIFF
--- a/src/minimega/misc_cli.go
+++ b/src/minimega/misc_cli.go
@@ -174,15 +174,21 @@ func cliRead(c *minicli.Command, respChan chan<- minicli.Responses) {
 			continue
 		}
 
+		var abort bool
+
 		for resp := range RunCommands(cmd) {
 			respChan <- resp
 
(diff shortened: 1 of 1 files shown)


jcrussell:
Now:

```
minimega$ .annotate false
minimega$ shell cat test
echo before
shell asdfasdfas
echo after
minimega$ read test
before
Error (foo): exec: "asdfasdfas": executable file not found in $PATH
```

djfritz:
There are more use cases where you want read to proceed - for example repeating an experiment that does things like shell commands that on the second run will return a false error, etc.

djfritz:
Closing re: discussion on #560

PR #562, opened 2016-06-07
minimega: cliPreprocessor improvements

jcrussell:
Prepend all commands from Namespace.Launch with the namespace. This
ensures that we invoke the cliPreprocessor.

Updates #555.

PR #577, opened 2016-06-14
minimega: make `file:` work with `cc send`

jcrussell:
Only prepend the files dir to the argument when it is not already in the
files dir. This allows the `file:` prefix to work with `cc send` so you
can push files from any node.


jcrussell:
Duplicate of #562.

PR #578, opened 2016-06-14
protonuke: tweaks

jcrussell:
Close smtp connections. Disable http(s) keep alives.


djfritz:
LGTM, though I wonder if we should tease out the keep-alive as a flag... merging anyways and we'll add that later if need be.

PR #582, opened 2016-06-23
UDP added, SSH and SMTP fixes

gmjohnson:
UDP added. Changes to protonuke.go and report.go to add UDP reporting/simulating. Works on AWS, however i got a weird disconnect error after running the simulation for a long time.

SSH - port changed to a non-standard port due to a port binding issue. Worked on AWS with non-standard port.

SMTP - added error checking to catch the io.EOF error, which was causing the connection to remain open and timout (also preventing new connections from opening). Additionally, changed the port to 587 because port 25 often filters SMTP traffic if it looks like spam (port 587 is used by spammers to bypass firewalls). Generates much more traffic and works on AWS.


src/protonuke/smtp.go
```diff
@@ -77,7 +79,7 @@ var (
 	max_size  = 131072
 	myFQDN    = "protonuke.local"
 	TLSconfig *tls.Config
-	smtpPort  = ":25"
+	smtpPort  = ":587" // nonstandard port used to evade blocking by a firewall
```
mwisely: We should probably put this back to `25`

src/protonuke/smtp.go
```diff
@@ -355,12 +359,14 @@ func (s *SMTPClientSession) Handler() {
 			case strings.HasPrefix(cmd, "RSET"):
 				s.mail_from = ""
 				s.rcpt_to = []string{}
-				s.addResponse("250 I forgot to remember to forget")
+				s.addResponse("250 I forgot to remember to forget")
```
mwisely: looks like these files need some `go fmt` cleanup

src/protonuke/smtp.go
```diff
@@ -355,12 +359,14 @@ func (s *SMTPClientSession) Handler() {
 			case strings.HasPrefix(cmd, "RSET"):
 				s.mail_from = ""
 				s.rcpt_to = []string{}
-				s.addResponse("250 I forgot to remember to forget")
+				s.addResponse("250 I forgot to remember to forget")			
 			default:
 				s.addResponse("500 unrecognized command")
+				break
```
mwisely: control will exit the switch without the break. it can be removed

src/protonuke/ssh.go
```diff
@@ -56,9 +56,10 @@ type sshConn struct {
 	StdoutBuf bytes.Buffer
 }
 
+// Non-default port used because there was a port binding issue with port 22
 var (
 	sshConns []*sshConn
-	PORT     = ":22"
+	PORT     = ":8080"
```
mwisely: Same as SMTP -- this should be the default port

src/protonuke/smtp.go
```diff
@@ -419,11 +424,16 @@ func (s *SMTPClientSession) readSmtp() (input string, err error) {
...
+				input = "QUIT"
+				err = nil
+			}
 			break
 		}
 		if strings.HasSuffix(input, suffix) {
 			break
 		}
 	}
+	log.Debugln("Read from the buffer")
```
djfritz: This is unnecessary.

src/protonuke/udp.go
```diff
@@ -0,0 +1,154 @@
...
+
+type udpConn struct {
+	Conn		net.Conn
+	Host		string
+
+}
+
+var (
+	udpConns []*udpConn
+	U_PORT	 = ":10001"
```
djfritz: Why port 10001?

djfritz:
The implementation looks fine, but udp is a transport, not a protocol. Should we be implementing a udp-based protocol instead?

src/protonuke/ssh.go
```diff
@@ -56,10 +56,9 @@ type sshConn struct {
 	StdoutBuf bytes.Buffer
 }
 
-// Non-default port used because there was a port binding issue with port 22
 var (
 	sshConns []*sshConn
-	PORT     = ":8080"
+	PORT     = ":22" // may be useful to use a non-standard port if ssh'd into a machine to avoid port binding issue
```
floren: You could make this an argument with a default of 22... although I dislike adding command line arguments, sometimes you gotta do it.

djfritz:
Other, bigger changes are coming to protonuke. We may revisit some of this eventually.

PR #587, opened 2016-06-30
protonuke: timeouts and ipv6

floren:
You can't do an Atoi on a hex address.

If you set an HTTP timeout of 30 seconds, it'll give up on a transfer after 30 seconds. I fixed that.


floren:
@djfritz @jcrussell

djfritz:
LGTM

PR #589, opened 2016-07-20
minimega: remove kill timeout a.k.a. I'll never let go

jcrussell:
If we have lots of containers and call `vm kill all` then we will likely
timeout when receiving the ACKs. Removed the timeout so that we'll
always wait until everything has ACK'ed.

There's probably a race condition if the VM quits while we are trying to
kill it.


djfritz:
LGTM

PR #592, opened 2016-07-22
minimega: gotta copy em' all

jcrussell:
Copy VMs before shoving them into the data field to prevent a race
condition on VM tags.


djfritz:
I don't merge pokemon themed PRs

djfritz:
LGTM

PR #594, opened 2016-07-22
minimega: make it work

jcrussell:
Pass flags to the shim so that CGROUP_ROOT is correct when you change the minimega base directory.

Also fix the locking with cloned VMs.

Getting close to booting...


floren:
LGTM

djfritz:
LGTM, tested well enough on a namespace-enabled experiment with thousands of endpoints and about 30 hosts.

PR #598, opened 2016-07-25
minimega: screenshots are now integral to vms

floren:
KVM VMs aren't the only type that may be able to do screenshots. Instead,
we can make Screenshot part of all VM types, and just have containers
return an error if you call it.

@djfritz @jcrussell

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -739,7 +739,7 @@ func (vm *ContainerVM) ConflictsContainer(vm2 *ContainerVM) error {
 }
 
 func (vm *ContainerVM) Screenshot(size int) ([]byte, error) {
-	return nil, errors.New("Cannot take screenshot of container")
+	return nil, errors.New("cannot take screenshot of container")
 }
 
 func (vm *ContainerConfig) String() string {
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -738,6 +738,10 @@ func (vm *ContainerVM) ConflictsContainer(vm2 *ContainerVM) error {
 	return vm.BaseVM.conflicts(vm2.BaseVM)
 }
 
+func (vm *ContainerVM) Screenshot(size int) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


src/minimega/vm_cli.go
```diff
@@ -723,10 +723,7 @@ func cliVmScreenshot(c *minicli.Command, resp *minicli.Response) error {
 		max = v
 	}
 
-	vm, err := vms.FindKvmVM(c.StringArgs["vm"])
-	if err != nil {
-		return err
-	}
+	vm := vms.FindVM(c.StringArgs["vm"])
```
djfritz: Should there be a nil check here?

floren: Absolutely yes, update incoming.

src/minimega/vm_cli.go
```diff
@@ -723,9 +723,9 @@ func cliVmScreenshot(c *minicli.Command, resp *minicli.Response) error {
 		max = v
 	}
 
-	vm, err := vms.FindKvmVM(c.StringArgs["vm"])
-	if err != nil {
-		return err
+	vm := vms.FindVM(c.StringArgs["vm"])
+	if vm == nil {
+		return errors.New("no such vm")
```
jcrussell: Use `func vmNotFound(idOrName string) error`

src/minimega/container.go
```diff
@@ -738,6 +738,10 @@ func (vm *ContainerVM) ConflictsContainer(vm2 *ContainerVM) error {
 	return vm.BaseVM.conflicts(vm2.BaseVM)
 }
 
+func (vm *ContainerVM) Screenshot(size int) ([]byte, error) {
+	return nil, errors.New("Cannot take screenshot of container")
```
jcrussell: Idiomatic errors are lowercase.

djfritz:
LGTM

PR #599, opened 2016-07-26
minimega: add VNC shim

jcrussell:
qemu now creates a domain socket for VNC connections to the VM. minimega creates a TCP<->domain socket shim that proxies all traffic using a random port for TCP. Updated new field in `vm info` and `vm summary` to lookup the port. Updated web to connect noVNC to the non-5900 based ports.

### DIFF
--- a/misc/web/js/glue.js
+++ b/misc/web/js/glue.js
@@ -55,7 +55,7 @@ function screenshotURL (vm, size) {
 
 // Generate the appropriate URL for requesting a VNC connection
 function vncURL (vm) {
-    return "./vnc#" + vm.host + ":" + (5900 + vm.id) + ":" + vm.name
+    return "./vnc#" + vm.host + ":" + (vm.vnc_port) + ":" + vm.name
 }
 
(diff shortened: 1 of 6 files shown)


floren:
LGTM

djfritz:
LGTM - can we confirm that this doesn't leak goroutines before merging? The debug API gives a count of running goroutines.

jcrussell:
No leaked goroutines from running:

```
#! /bin/bash

for i in $(seq 5); do
    bin/minimega -e "vm launch kvm 1"
    bin/minimega -e "vm start all"
    port=$(bin/minimega -e ".annotate false .headers false .columns vnc_port vm info")

    for j in $(seq 5); do
        timeout 10s vinagre :$port
    done

    bin/minimega -e "vm kill all"
    bin/minimega -e "vm flush"

    num=$(bin/minimega -e ".annotate false .headers false .columns Goroutines debug")

    echo "$(date): $num" >> goroutines.log
done
```

@djfritz

PR #602, opened 2016-07-27
docs: add more cgroup setup info to documents

floren:
Give concrete example of how to turn on the memory cgroup for Debian.


djfritz:
LGTM

PR #603, opened 2016-07-27
minimega: add -ltermcap to goreadline LDFLAGS

floren:
Although it's not needed on Debian, some distros (e.g. Slackware) appear to
need this. Tested that it still compiles on Debian.

### DIFF
--- a/src/goreadline/goreadline.go
+++ b/src/goreadline/goreadline.go
@@ -5,7 +5,7 @@
 // Very simple libreadline binding.
 package goreadline
 
-// #cgo LDFLAGS: -lreadline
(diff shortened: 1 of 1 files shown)


floren:
@djfritz

djfritz:
LGTM, does this still run on debian/ubuntu (I can totally see systemd enabled debian requiring that the flag NOT be there or some silliness)?

mwisely:
Seems to work fine on Ubuntu 14.04, for whatever that's worth

jcrussell:
LGTM. Also works on Fedora 22.

PR #605, opened 2016-07-28
goreadline/minimega: sigint like bash

jcrussell:
Respond to Ctrl-C like bash: kill the current line and start with a new prompt.

Only thing that is different is that the reverse history line gets cleared when searching is aborted by Ctrl-C.


src/goreadline/goreadline.c
```diff
@@ -1,11 +1,85 @@
-// Copyright (2015) Sandia Corporation.
```
djfritz: Why did we lose the copyright header?

src/goreadline/goreadline.go
```diff
@@ -84,13 +91,13 @@ func minicliCompletion(text *C.char, start, end C.int) **C.char {
 // Rlwrap prompts the user with the given prompt string and calls the
 // underlying readline function. If the input stream closes, Rlwrap returns an
 // EOF error.
-func Rlwrap(prompt string, record bool) (string, error) {
+func Readline(prompt string, record bool) (string, error) {
```
djfritz: Function name changed, but function name in the comment didn't.

src/goreadline/goreadline.c
jcrussell: Whoops, deleted that whole file and forgot to re-add the header.

src/goreadline/goreadline.c
```diff
@@ -1,11 +1,87 @@
...
-// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
```
djfritz: What about the rest of the header?

djfritz:
LGTM

Issue #556, opened 2016-06-06
protonuke: generate DNS traffic


jcrussell:
Generate DNS traffic independently of other traffic generation. Possibly using [0] or gopacket.

[0] https://github.com/miekg/dns

PR #609, opened 2016-07-29
protonuke dns implementation

cdshann:
This addresses #556 . 

Client:
* Makes DNS requests for top 100 domains in domains.go
* Flag dnsv4 will instruct the client to only make A record requests
* Flag dnsv6 will instruct the client to only make AAAA record requests
* Default is for the client to randomly select between A, AAAA, CNAME, MX, NS, and SOA records.

Server:
* Implements all records requested by the client
* For A/AAAA records, the server will first attempt to getRandomHost() and see if a compatabile address was specified in the targets parameter. If a suitable address is not found one is randomly generated.
* For all other records the server alters the original question section
* Defaults to UDP, but TCP is supported by the DNS library


djfritz:
Somebody wants a LOC bump...

djfritz:
This looks fine. How much has this been tested?

cdshann:
I just finished testing all possible protonuke<->protonuke interactions. I also verified the server was responding correctly to the implemented question types with Dig.

jcrussell:
Can you `cp src/dns/COPYRIGHT LICENSES/LICENSE.dns`?

cdshann:
Done.

A question about the DNS server - would you rather a server configured without an IPv6 (or IPv4) hosts range reply with an NXDOMAIN response instead of a randomly generated IP when receiving a AAAA/A record request?

jcrussell:
We decided to add yet another flag -- `-random-hosts` and return `NXDOMAIN` if hosts is empty and `-random-hosts` is not set.

jcrussell:
LGTM

@djfritz: merge?

djfritz:
LGTM

PR #610, opened 2016-08-02
Meshage fix

floren:
Minimega used to only broadcast on one interface. Now, by default, it broadcasts on all ipv4 interfaces. If you specify "-interfaces eth0,eth1" it will only broadcast on eth0 and eth1.

@djfritz @jcrussell

### DIFF
--- a/src/meshage/node.go
+++ b/src/meshage/node.go
@@ -80,6 +80,8 @@ type Node struct {
 	updateNetwork    bool
 	Snoop            func(m *Message)
 	version          string
+
+	broadcastAddrs	[]*net.UDPAddr	// list of addresses to use as SOURCE addresses for UDP broadcasts
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


djfritz:
Looks like we enumerate broadcastAddrs once in NewNode(). What happens if the interface changes IP? Does it still work?

floren:
I was just going to say "don't do that" but we can check at every broadcast
if needed. I wanted to avoid checking every single tap we create for vms,
but your point about changing ips makes sense

On Aug 3, 2016 3:16 PM, "David Fritz" <notifications@github.com> wrote:

Looks like we enumerate broadcastAddrs once in NewNode(). What happens if
the interface changes IP? Does it still work?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/sandia-minimega/minimega/pull/610#issuecomment-237376395>,
or mute the thread

djfritz:
What's the status on this? If the discussion is dead can we close this?

floren:
Depends if we care about this functionality or not :)

PR #614, opened 2016-08-10
Change mkfs.ext4 to mkfs.ext3

floren:
because syslinux and ext4 don't get along well in all situations. For the time being, ext3 should be safe in all cases.


djfritz:
LGTM

Issue #611, opened 2016-08-04
VNC 'connect' button in web not getting correct URL


glattercj:
As of PR #599, the 'connect' button in the web is showing 'undefined' for the vnc_port in the URL.

I can see the snapshot, but when I click on the connect button, the url says:

`http://localhost:9001/vnc/#halo9:undefined:subnet_0.server_1`

djfritz:
confirmed and #599 broke the build anyway, so I reverted it. @jcrussell is a monster.

jcrussell:
Pretty sure this worked for me when I tested it... did @glattercj pull the latest misc/ changes?

djfritz:
Either way that pr doesn't compile - missing package import in kvm.go or
something

jcrussell:
Odd... it compiles just fine in my branch but not when merged?

djfritz:
Maybe another pr that merged first removed a stale import?

jcrussell:
I just did the merge locally and it compiles. Will look at the 'connect' button issue.

glattercj:
@jcrussell 
I cloned the master branch of this repo...I'm assuming that would include any misc/ changes.

jcrussell:
Merging `origin/master` reverted my code on my branch... d'oh. Will have to look at this next week, don't have time today.

jcrussell:
@glattercj: are the connect URLs wrong still?

glattercj:
@jcrussell Yes we updated to commit 1a80e6ba7ef9310832ab58115e0e56477fbe1ce9.

Yes, the URLs are still wrong: `http://localhost:8999/vnc/#compute10:undefined:foobar`

Was this supposed to be fixed in #617?

jcrussell:
@glattercj: I haven't seen this error when testing on CCC. Did you update minimega on `compute10`?

glattercj:
Nevermind...I was looking at 'git log' for the commit, so I must have pulled the fix, but never actually built it. After building, the VNC links work like they should. Thanks @jcrussell.

PR #617, opened 2016-08-15
minimega: retry #599

jcrussell:
Fixed import issue that somehow occurred when we merged #599. Reverts #612.

Wasn't able to duplicate #611.


djfritz:
LGTM

PR #623, opened 2016-08-19
minimega: implement token bucket for launching vms

jcrussell:
Added rate/burst parameters to control how fast KVM VMs get launched.
This allows KSM to keep up.

### DIFF
--- a/src/minimega/main.go
+++ b/src/minimega/main.go
@@ -21,7 +21,6 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
-	"time"
 	"version"
 )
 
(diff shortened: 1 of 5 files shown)


src/minimega/main.go
```diff
@@ -41,6 +42,8 @@ var (
 	f_msaTimeout = flag.Uint("msa", 10, "meshage MSA timeout")
 	f_port       = flag.Int("port", 9000, "meshage port to listen on")
 	f_ccPort     = flag.Int("ccport", 9002, "cc port to listen on")
+	f_vmRate     = flag.Duration("vm-rate", time.Second/4, "average rate to launch KVM VMs")
```
djfritz: this should be in the api instead of a flag - if i boot something and it fails, i'd like to change the metrics without having to restart minimega.

jcrussell:
Should we add `vm rate [rate]` and `vm burst [burst]` APIs?

jcrussell:
We should probably do some testing to find out what are reasonable default values. Currently, the rate is 0.25 second and the burst limit is 10.

djfritz:
Still unclear if we actually need this. Waiting for now.

djfritz:
Looks like things work okay without this, and this is something the new scheduler should be taking care of. Closing this for now; we can resurrect this topic when the scheduler PR comes up.

PR #626, opened 2016-08-24
minimega: `make it weird for everyone`

jcrussell:
Add container fields to `vm info` columns. Reorder things so that they
make some sense... also update `vm summary` to mirror the same order.


djfritz:
LGTM

PR #628, opened 2016-08-24
minimega/minicli/ranges: moving code

jcrussell:
Moving `compressHosts` function from minicli to ranges as `UnsplitList`.
This function was previously used by minicli's output formatter to
condense a list of hostnames into a pretty range. We now use this when
printing the hosts that belong to a namespace.


djfritz:
LGTM

PR #645, opened 2016-08-26
minimega: fix vm config net help

### LINKED ISSUE #643: Trivial: Update docs for vm net
```help vm config net```
```
These aliases will be allocated from the pool of available VLANs and is
namespace-aware (i.e. 'DMZ' in namespace 'foo' will be a different VLAN than
'DMZ' in namespace 'foo'). Internally, this is implemented by concatenating the
namespace name with the VLAN alias (e.g. 'DMZ' in namespace 'foo' becomes
'foo//DMZ'). If you wish to connect VLANs in different namespaces, you may use
abuse this implementation detail:

        namespace bar
        vm config net foo//DMZ

Calling vm net with no parameters will list the current networks for this VM.
```
I think you mean the VLANs will be the same and you mean to use foo and bar

Also "use abuse" should be "use"

"Calling vm net " should be "Calling vm config net"

jcrussell:
Nice catch kunz.

Fixes #643


djfritz:
LGTM

Issue #644, opened 2016-08-26
minimega: `cc proccess` improvements


jcrussell:
It would be better if `cc process list` were based on`cc filter` rather than the `<vm id, name, uuid or all>` argument.

`cc process kill <pid or all>` is a bit inconvenient since you have to look up the PID first before killing something. Perhaps instead, we should store a counter in ron/client.go and kill based on the counter value rather than the actual PID. Alternative, we could also support something like `pkill` or `killall`.

jcrussell:
Rolling this into #666.

djfritz:
captured in #666

PR #646, opened 2016-08-29
minimega/ron: add killall functionality

jcrussell:
KillAll processes by name. Updates #644.

Should work... need to test though.


djfritz:
Should this match on filepath.Base() instead of strings.Contains()?

jcrussell:
I thought about that a bit... what if the command you wanted to kill was `bash run.sh`? Then you'd have to `pkill bash` which would be no good. `killall run.sh` would work with the current implementation.

I think the real killall does what you suggest:

```
Killing by file only works for executables that are kept open during execution,
i.e. impure executables can't be killed this way.
```

djfritz:
LGTM

PR #654, opened 2016-09-08
minimega: fix concurrent map access

### LINKED ISSUE #653: minimega: race with vm tags
Pretty sure that the issue @floren and I have been looking into is a race with vm tags. I've seen two stack traces now (one involving SetTags, and the other gob) but I didn't think to save the first. I modified go's runtime to dump the type of the map with the concurrent access and it's a `map[string]string` -- which could be the tags.

It would make sense with #565 that there could be an issue when minirouter updates the tags while we're running `vm info`.

jcrussell:
Need to copy the BaseConfig which contains the VM's tags in the VM Copy
functions. Fixes #653. Tested with 20k curls to localhost:9001/vms/.

@floren

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -608,6 +608,7 @@ func (vm *ContainerVM) Copy() VM {
 	defer vm2.lock.Unlock()
 
 	// Make deep copies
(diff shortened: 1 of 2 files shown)


floren:
LGTM

PR #655, opened 2016-09-08
protonuke: smarter file size parsing

jcrussell:
Parse the file size as a custom flag type so that we can have suffixes
like B/KB/MB. Default is still MB but now we can specify images smaller
than 1MB using the B or KB suffix.


djfritz:
Fancy!

LGTM

PR #656, opened 2016-09-08
nfcat: tool to read binary netflow

jcrussell:
Adding tool `nfcat` that reads the binary netflow format written by minimega.

Required small changes to gonetflow which haven't been tested yet... probably fine?


djfritz:
LGTM

PR #660, opened 2016-09-14
minimega: add uptime

jcrussell:
We're so stable, we need an uptime API.


djfritz:
Here's to this reducing stability.

LGTM

PR #662, opened 2016-09-14
protonuke: improvements

jcrussell:
Three improvements:
 * add `-httpgzip` flag
 * return 202 on POST
 * use stdlib to parse hosts (and remove broadcast host)
 * remove second 0 from alpanum


jcrussell:
In pieces... almost ready to roll it into latest experiment.

jcrussell:
This is well tested at this point.

PR #663, opened 2016-09-15
minimega/minirouter: add upstream dns

jcrussell:
Add option to set upstream dns server. Useful when creating a hierarchy.

### DIFF
--- a/src/minirouter/dnsmasq.go
+++ b/src/minirouter/dnsmasq.go
@@ -13,9 +13,10 @@ const (
 )
 
 type Dnsmasq struct {
-	DHCP map[string]*Dhcp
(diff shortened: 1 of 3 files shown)


src/minirouter/dnsmasq.go
```diff
@@ -163,6 +169,10 @@ var dnsmasqTmpl = `
 # don't read /etc/resolv.conf
 no-resolv
 
+{{ if ne .Upstream "" }}
+	server={{.Upstream}}
```
djfritz: Don't we want the local option always enumerated here?

djfritz: I'm a potato - this works as advertised.

PR #668, opened 2016-09-20
minimega: dnsmasq needs to run as root

floren:
If dnsmasq defaults to running as "nobody", it won't be able to access its directory under /tmp/minimega. This fixes that.


floren:
@djfritz @jcrussell

Issue #666, opened 2016-09-16
multiple: rewrite ron/minitunnel


djfritz:
ron is doing too much and doing it poorly at that. Refactor to move command/response code to minimega, transport/redial/tunnels/serial/UDS stays in ron, and command processing, etc. goes to miniccc.

jcrussell:
Closing this now that we've beat the devil out of it.

PR #675, opened 2016-09-27
minimega: fix deadlock with flushing VMs

jcrussell:
Call to ron.Server.UnregisterVM while holding VM's lock could result in
a deadlock. Moved call to unregister VM to VMs, mirroring how RegisterVM
is called.


jcrussell:
This replaces a rare deadlock with a more frequent deadlock -- we need to work on #666 soon.

Issue #426, opened 2016-02-26
minimega: extend VNC API


floren:
The VNC API could use an overhaul. Here's Fritz's suggestion:

```
vnc fb record <host> <vm> <filename>
vnc kb record <host> <vm> <filename>
vnc fb stop <host> <vm>
vnc kb stop <host> <vm>

# show current recordings/playbacks. More info is needed at least on playbacks (time remaining?, % complete...) as well as recordings (time recorded...)
vnc     

# start a playback or restart from the beginning of an existing playback
vnc play <host> <vm> <filename>

# stop playback, but keep the current position in the playback file
vnc stop <host> <vm>

# continue a stopped playback
vnc continue <host> <vm>

# if started, skip the current delay and advance immediately to the next event (@floren's need)
# if stopped, simply play the next event and stop again
vnc step <host> <vm>
```

floren:
@cdshann you want this?

djfritz:
While we're at it, let's add inject capability.

    vnc inject <host> <vm> <kb,mouse> ...

cdshann:
@floren If this can wait until the middle of next week for me to start (need to finish nuke fixes and wrap up qos) then yeah I will take it. Thanks

mkunz7:
Instead of "stop" can we use the word "pause". I think stop implies it resets to the start. 

Floren's need was actually by my request. Sorry to create so much work.

To get a clearer understanding of my need. In some of the complex scripts I write I press next to start an installer then wait for an hour for the installer to finish.

100000000:KeyEvent,true,Alt_L
100000000:KeyEvent,true,n
100000000:KeyEvent,false,n
100000000:KeyEvent,false,Alt_L
3600000000000:PointerEvent,0,0,0
100000000:KeyEvent,true,Alt_L
100000000:KeyEvent,true,f
100000000:KeyEvent,false,f
100000000:KeyEvent,false,Alt_L

When writing these scripts I've been padding them so if the installer takes 20min I give it an hour. This is so we when move between hardware it has a greater chance of working. When debugging I have to sit and wait that extra 40min and with Floren's new skip or step command whatever you want to call it I can tell it to stop waiting.

I can see the value add with single stepping where if I had a better way to determine what is on the screen things can be live.

Additional request with step. Make it so a VNC doesn't have to be running and I can issue a step.
vnc step <host> <vm> 100000000:PointerEvent:0,0,0
vnc step <host> <vm> 100000000:PointerEvent:1,0,0
[...]

PR #513, opened 2016-04-27
minimega: vnc api extensions

cdshann:
This is a refactor of the minimega vnc recording and playback code to support the features requested in #426.

This is not yet ready to merge, but I wanted to create the PR to get any feedback and so everyone could track the development.

PR #676, opened 2016-09-27
vnc namespaces

cdshann:
This PR makes the VNC api namespace aware, and also moves the vnc recordings onto the host where the vm is running. 

Updates #426 .


src/minimega/vnc_cli.go
```diff
@@ -27,68 +28,101 @@ If playback is selected, the specified file (created using vnc record) will be
...
 		HelpShort: "reset VNC state",
 		HelpLong: `
 Resets the state for VNC recordings. See "help vnc" for more information.`,
 		Patterns: []string{
 			"clear vnc",
 		},
-		Call: wrapSimpleCLI(func(_ *minicli.Command, _ *minicli.Response) error {
-			return vncClear()
-		}),
+		Call: wrapVMTargetCLI(cliVNCClear),
```
jcrussell: Change to `wrapBroadcastCLI`.

src/minimega/vnc_cli.go
```diff
@@ -27,68 +28,101 @@ If playback is selected, the specified file (created using vnc record) will be
...
+List all running vnc playback/recording instances. See "help vnc" for more information.`,
+		Patterns: []string{
+			"vnc",
+		},
+		Call: wrapBroadcastCLI(vncList),
 	},
 }
 
+// List all active recordings and playbacks
+func vncList(c *minicli.Command, resp *minicli.Response) error {
```
jcrussell: Rename to `cliVNCList`. Should only list if it matches the active namespace.

src/minimega/vnc.go
```diff
@@ -61,37 +60,21 @@ func init() {
...
+		return nil, vmNotFound(name)
 	}
 
-	rhost := fmt.Sprintf("%v:%v", host, vm.VNCPort)
+	rhost := fmt.Sprintf("%v:%v", name, vm.VNCPort)
 
 	c := &vncClient{
 		Rhost: rhost,
-		Host:  host,
+		Host:  hostname,
```
jcrussell: Might be worth adding a reference to the VM to the `vncClient` struct. Then you don't need to store Name or ID.

src/minimega/vnc_cli.go
```diff
@@ -103,36 +137,37 @@ func cliVNC(c *minicli.Command, resp *minicli.Response) error {
...
-		return vncPlaybackKB(host, vm, fname)
+		return vncPlaybackKB(vm, fname)
 	}
+	return nil
+}
 
-	// List all active recordings and playbacks
-	resp.Header = []string{"host", "name", "id", "type", "filename"}
-	resp.Tabular = [][]string{}
+func cliVNCClear(c *minicli.Command, resp *minicli.Response) error {
```
jcrussell: Should we do something akin to `cliVNCClear` when a VM is flushed? What cleans up these maps?

src/minimega/tunnel.go
```diff
@@ -17,79 +17,70 @@ import (
...
-					if !strings.Contains(err.Error(), "closed network connection") && err != io.EOF {
-						log.Errorln(err)
-					}
-					break
-				}
-
-				err = websocket.Message.Send(ws, rbuf[:n])
-				if err != nil {
+		rbuf := make([]byte, VNC_WS_BUF)
+		for {
```
jcrussell: This *seems* unnecessary... we're already copying from remote to ws above. Should figure this out.

src/minimega/vnc.go
```diff
@@ -26,9 +25,8 @@ var (
 )
 
 type vncClient struct {
-	Host  string
-	Name  string
-	ID    int
+	Vm    *KvmVM
+	Ns    string
```
jcrussell: `Vm` -> `VM`. `Ns` -> `ID // ID is namespace:name`.

src/minimega/vnc_cli.go
```diff
@@ -27,72 +28,106 @@ If playback is selected, the specified file (created using vnc record) will be
...
 		Patterns: []string{
-			"vnc",
-			"vnc <kb,fb> <record,> <host> <vm id or name> <filename>",
-			"vnc <kb,fb> <norecord,> <host> <vm id or name>",
-			"vnc <playback,> <host> <vm id or name> <filename>",
-			"vnc <noplayback,> <host> <vm id or name>",
+			"vnc <kb,fb> <record,> <vm name> <filename>",
+			"vnc <kb,fb> <norecord,> <vm name>",
+			"vnc <playback,> <vm name> <filename>",
+			"vnc <noplayback,> <vm name>",
```
jcrussell: While we're changing things... should we swap the order of `<kb,fb>` and `<record>`?

src/minimega/vnc_cli.go
```diff
@@ -27,72 +28,106 @@ If playback is selected, the specified file (created using vnc record) will be
...
-		host = hostname
+	vm, err := vms.FindKvmVM(c.StringArgs["vm"])
+	if err != nil {
+		return fmt.Errorf("vm %s not found", c.StringArgs["vm"])
 	}
 
 	if c.BoolArgs["record"] && c.BoolArgs["kb"] {
 		// Starting keyboard recording
-		return vncRecordKB(host, vm, fname)
+		return vncRecordKB(vm.Name, fname)
```
jcrussell: We already have a reference to the VM -- we should pass it in here and then we don't have to find it again later in NewVNCClient.

src/minimega/vnc_cli.go
```diff
@@ -27,72 +28,106 @@ If playback is selected, the specified file (created using vnc record) will be
...
 	} else if c.BoolArgs["record"] && c.BoolArgs["fb"] {
 		// Starting framebuffer recording
-		return vncRecordFB(host, vm, fname)
+		return vncRecordFB(vm.Name, fname)
```
jcrussell: Same as above.

src/minimega/vnc_cli.go
```diff
@@ -27,72 +28,106 @@ If playback is selected, the specified file (created using vnc record) will be
...
-			for k, v := range vncKBRecording {
-				if v.Matches(host, vm) {
-					client = v.vncClient
-					delete(vncKBRecording, k)
-					break
-				}
+			err = fmt.Errorf("kb recording %v not found", vm.Name)
+			if v, ok := vncKBRecording[ns]; ok {
+				client = v.vncClient
+				delete(vncKBRecording, ns)
```
jcrussell: Is this the right place to delete from `vncKBRecording`? Need to think about this.

src/minimega/vnc_cli.go
jcrussell: Ah, now I see what you did here. I would move the current `cliVNCClear` code into `vncClear` and change the `clear vnc` handler to:

```
        Call: wrapBroadcastCLI(func(_ *minicli.Command, _ *minicli.Response) error {
            vncClear()
            return nil
        }),
```

src/minimega/kvm.go
```diff
@@ -428,6 +429,7 @@ func (vm *KvmVM) connectVNC() error {
 	// Keep track of shim so that we can close it later
 	vm.vncShim = l
 	vm.VNCPort = l.Addr().(*net.TCPAddr).Port
+	ns := fmt.Sprintf("%v:%v", GetNamespaceName(), vm.Name)
```
jcrussell: This should be the VM's name, not the active namespace.

jcrussell:
We're going to punt on the issue of adding a sync.RWLock for the vnc* maps. Will be addressed in #513.

djfritz:
Will merge after the branch gets fixed.

cdshann:
Don't merge me yet.

PR #685, opened 2016-10-05
powerbot: Remove an unused source file from powerbot

mwisely:
`ssh.go` isn't used anywhere in powerbot. It'll build without it.


djfritz:
How on earth did we hit a merge conflict with this... Will merge after it gets updated.

mwisely:
Hmm. I'm on it.

mwisely:
@djfritz I think it's good now. I'm not sure what happened.

jcrussell:
FWIW, the merge conflict was due to changes in the way we vendorized code.

PR #688, opened 2016-10-11
add a web console for containers

floren:
Import xterm.js to provide a javascript-based console. The minimega web
    interface now provides a "Connect" button for containers, which will
    connect to a javascript terminal.
    
    The terminal is implemented as a pseudo-terminal, rather than with
    three pipes as used previously. If we regularly plan to use thousands
    of containers on the same host, we will need to include code in minimega
    startup to increase the number of PTYs available on the system.
    
    Currently, this terminal behaves exactly as the domain socket: if you
    open the terminal, then close it and re-open, the previous output will
    not be re-displayed. This will be implemented later.


floren:
@jcrussell @djfritz

jcrussell:
Looks pretty good overall. A few nits:

 * Can you move the `pty` code to the vendor tree?
 * Can you symlink the LICENSE into LICENSES/
 * Same for xterm's license.
 * Should `terminalWsHandler` match `vncWsHandler`? The latter is a bit cleaner, and...
 * Can you change the URL for the terminal to match `/vnc/`? `/terminal/<name>` is more friendly than supplying a host/port.

floren:
Checking them off one by one :)

jcrussell:
Looks like we will keep listening, even after the container is flushed. In `KvmVM`, we keep a handle to the listener (`vncShim`) which we close after the qemu process exits.

PR #699, opened 2016-10-14
misc: update ccc host image

jcrussell:
Include netcat-openbsd instead of regular netcat. Add curl.


djfritz:
It may be the other way around:

    foo@foo:~$ nc -U
    This is nc from the netcat-openbsd package. An alternative nc is available
    in the netcat-traditional package.
    usage: nc [-46bCDdhjklnrStUuvZz] [-I length] [-i interval] [-O length]
              [-P proxy_username] [-p source_port] [-q seconds] [-s source]
              [-T toskeyword] [-V rtable] [-w timeout] [-X proxy_protocol]
              [-x proxy_address[:port]] [destination] [port]

djfritz:
Did we confirm which version of netcat we really want?

jcrussell:
I think we want netcat-traditional... I'll rebuild and test.

floren:
netcat-openbsd is what we want. LGTM.

Issue #712, opened 2016-10-20
minimega: `cc clients` is not filtered by namespace


jcrussell:
And it should be.

jcrussell:
Fixed in #714.

Issue #713, opened 2016-10-21
minimega: env preprocessor


jcrussell:
It would be trivial to add a `$` preprocessor that looks for a variable in env. For example:

```
vm config filesystem $containerfs
```

This would allow a lot more flexibility in the minitest cases as you could setup the required env variable to point to where the resources are locally and then run the same tests everywhere.

PR #714, opened 2016-10-21
ron/miniccc/minimega: more refactoring

### LINKED ISSUE #710: minimega: `clear cc responses` should be namespace-aware
Currently deletes all responses.

jcrussell:
Improve cc compatibility with namespaces. Added a new and comprehensive
test of interaction between cc and namespaces that we can now pass.

Fixes #710, #712.

Need to do a bigger test of this.

### DIFF
--- a/src/miniccc/client.go
+++ b/src/miniccc/client.go
@@ -17,15 +17,11 @@ import (
 	"version"
 )
 
-var Client struct {
+var client struct {
 	ron.Client // embed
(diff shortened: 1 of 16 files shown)


jcrussell:
Note: location of containerfs is hardcoded in the test to where it is on my machine. Not sure how to get around this unless we implement #713.

jcrussell:
Tested this an it looks like it's working well.

@djfritz

PR #726, opened 2016-10-25
vnc injection

cdshann:
Per our discussion at the last minimeeting:
- Added the ability to inject vnc commands even if there is no playback currently running.
- Added the ability to inject LoadFile events from the cli. If there is no playback currently running a new vnc playback will be created.

### DIFF
--- a/doc/content/articles/vnc.article
+++ b/doc/content/articles/vnc.article
@@ -1,7 +1,6 @@
 VNC recording and playback
 
-The minimega authors
-11 May 2015
+The minimega authors 28 October 2016
 
 * Introduction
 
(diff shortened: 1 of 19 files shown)


jcrussell:
Looks good overall. Anything you want to throw into the minitest?

jcrussell:
Looks like the tests need fixing:

```
ERROR command_socket.go:97: invalid command: `vnc kb record vm0 /tmp/vm0.kbr`
```

doc/content/articles/vnc.article
```diff
@@ -62,8 +63,8 @@ To stop recording:
 
 The recorded file format uses the following schema:
 
-	<time delta>:PointerEvent,<mask>,<x>,<y>
```
floren: These should probably remain tabs instead of spaces... is somebody using emacs?

PR #727, opened 2016-10-25
minimega: make capture more namespace aware

### LINKED ISSUE #707: minimega: `capture pcap vm <vm> ...` should be namespace-aware
It's hard to make the whole API namespace-aware but at least the vm part should be.

jcrussell:
Made the capture API more namespace aware, still not fully though. Fixed
a bug in `clear capture` where it was only stopping the first capture.
Passes local test, need to try in a multiple node experiment.

Also: fix segfault in gopcap and minor improvement to minicli.

Fixes #707

@glattercj -- do you want to test this?

### DIFF
--- a/src/gopcap/gopcap.go
+++ b/src/gopcap/gopcap.go
@@ -22,12 +22,15 @@ import (
 type Pcap struct {
 	handle       unsafe.Pointer
 	dumperHandle unsafe.Pointer
+	done         chan bool
 	lock         sync.Mutex
 }
 
 // NewLearner returns an IPMacLearner object bound to a particular interface.
 func NewPCAP(dev string, file string) (*Pcap, error) {
-	ret := &Pcap{}
+	ret := &Pcap{
+		done: make(chan bool),
+	}
 
 	devC := C.CString(dev)
 	fileC := C.CString(file)
@@ -48,7 +51,10 @@ func NewPCAP(dev string, file string) (*Pcap, error) {
 	}
 	ret.dumperHandle = unsafe.Pointer(dumperHandle)
 
-	go C.gopcapCapture(handle, dumperHandle)
+	go func() {
+		C.gopcapCapture(handle, dumperHandle)
+		close(ret.done)
+	}()
 
 	return ret, nil
 }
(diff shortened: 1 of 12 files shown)


glattercj:
Yea sure.

glattercj:
I am able to capture successfully via a namespace, but I'm not able to delete captures individually. Under the main help, it says `capture pcap delete <id>`, but that doesn't work with namespaces. Under the namespace specific notes, it says you can clear captures across a namespace with `clear capture`. Does this mean I can't stop individual captures with namespaces?

jcrussell:
The problem is that captures are identified by ID which isn't globally unique across multiple minimega instances. If you said `capture pcap delete 0`, what instance should that run on? Multiple nodes may have a capture with ID 0.

One solution would be to run `capture pcap` and parse through the results to find the host and capture ID for the VM you wish to stop capturing. You could then `mesh send <host> capture pcap delete <ID>`. This moves the work to whatever you're using to drive minimega which isn't ideal.

Another solution would be to have a VM-specific delete. In general, we don't like adding arguments to the `clear` APIs -- `clear capture vm <name>` would not be idiomatic. I don't really like any of the following though:

```
capture pcap delete vm <name>
capture delete vm <name>
capture vm delete <name>
```

@djfritz -- any suggestions?

jcrussell:
Maybe we can change the `capture` API slightly -- we already assume that there is only one capture per bridge. If we make this explicit, we could support the following delete APIs:

```
capture pcap delete bridge <bridge>
capture pcap delete vm <name>
capture netflow delete bridge <bridge>
```

And do away with the IDs entirely.

jcrussell:
@glattercj -- want to take another look at this?

glattercj:
@jcrussell @floren 
- capture works
- `delete vm all` works
- `delete vm <name>` works (it *does* delete the capture), but I get an error message:
```
     $ minimega -e namespace foo capture pcap delete vm barbaz
     Error (dev): a capture for vm barbaz does not exist
```

floren:
@glattercj can you either post here or email me the exact commands you used to set up VMs and captures etc. for that test? I haven't seen that error message at all.

jcrussell:
Ah, this is an artifact of `wrapVMTargetCLI` -- @floren: you should return a `vmNotFound` error when a VM is not found rather than the error `a capture for vm barbaz does not exist`.

floren:
Yeah, Casey and I just figured that out, change incoming.

floren:
PTAL

jcrussell:
I want to make a few tweaks to the code first.

djfritz:
Still LGTM

PR #736, opened 2016-10-31
web: Interface enhancements

jasoncontrib:
Added buttons to customize column visibility in tables. Added additional columns, and removed ID column. Separate functionality out into separate pages so users can open different views in different browser tabs / windows easily. Fix the search functionality so that all table columns are searched properly. Minimize / refactor the front-end JavaScript to use built-in DataTables functionality where possible.


src/minimega/web.go
```diff
@@ -201,16 +207,48 @@ func webScreenshot(w http.ResponseWriter, r *http.Request) {
...
-func webTileVNC(w http.ResponseWriter, r *http.Request) {
-	http.ServeFile(w, r, filepath.Join(web.Root, "tilevnc.html"))
+// Templated HTML responses
+func webTemplated(w http.ResponseWriter, r *http.Request) {
+	lp := filepath.Join(web.Root, "templates", "_layout.tmpl")
+	fp := filepath.Join(web.Root, "templates", r.URL.Path+".tmpl")
+
+	info, err := os.Stat(fp)
+	if err != nil {
+		if os.IsNotExist(err) {
```
jcrussell: You can get rid of `if err != nil`: `os.IsNotExist(err)` returns false if `err` is nil. I would probably return 404 if `err != nil`.

src/minimega/stats.go
```diff
@@ -108,6 +114,24 @@ func cliHost(c *minicli.Command, resp *minicli.Response) error {
 	return nil
 }
 
+func hostStatsUptime() (string, error) {
```
jcrussell: Hah, uptime before was how long minimega has been running. Are both useful? I was thinking we were going to kill the `uptime` API and report how long minimega has been running in `host uptime` but maybe that's confusing.

@djfritz

jcrussell:
Any objections to merging this?

@floren @djfritz

floren:
hit it!

djfritz:
LGTM

Issue #536, opened 2016-05-19
More refined scheduling using namespaces


derhart:
Where on your roadmap is the improvements to scheduling in namespace-enabled minimega? Round-robin is working for us at the moment, but I can foresee us needing a more intelligent approach soon. We can implement a new approach, but I don't want to duplicate effort.

jcrussell:
I don't plan to touch the scheduler for a while -- I'd like to work the bugs out of master and (probably) cut a 2.3 release towards the end of summer. The main lift for 2.4 could be an improved scheduler with a shorter lead time.

What issues are you having with the round-robin scheduler? Perhaps we can make it marginally smarter to suit your needs.

djfritz:
Just echoing @jcrussell - we intentionally decoupled the scheduler from the namespace functionality, and left round-robin as a placeholder. This has come up a few times and we have a good approach for the solution when we eventually get to it. The earliest I can see this happening is August.

derhart:
I'm not sure it's a pressing issue. If you weren't going to look at it until CY17, I might be worried. Can you give me a clue of what your approach will be? Personally I'd like something that accounts for host "load" (however that's calculated) to determine the number of VM instances that can run on the host. That's about as advanced as I'd like. That way if someone short-circuits the typical namespace/meshage deploy strategy and deploys 90 VMs to a single host, any subsequent deployments will most likely ignore the overloaded host.

jcrussell:
Yes, we will try to schedule VMs according to the host "load". However, determining "load" is challenging -- VMs are started in a paused state so we can't know how much memory/CPU they will use when they actually start and the experiment begins (and how things like Kernel Samepage Merging will behave).

djfritz:
There's lot of research on VM placement and non-runtime scheduling (I'm opposed to the mess involved in runtime rescheduling). You can get very good solutions with a multi-level queue scheduler with each level scheduling on one constraint (memory, cpu load, similar vm type, similar network type). It's easy to tune as well to get most any type of schedule you want. 

I assume we'll have a schedule api somewhere to tune that or override with static scheduling (need vm x on node y).

djfritz:
@jcrussell 

A proposed scheduler:

Assumptions:

* there are N physical nodes
* there are X vms to schedule
  * a vm has one or more vlans
  * a vm has a disk or filesystem associated with it
  * a vm has other things, but they generally don't matter
* we don't need to use all physical hosts in a schedule, even if X > N, however:
[...]

PR #739, opened 2016-11-01
minimega: implement smarter scheduler

jcrussell:
Implements a least loaded first scheduler. Added two new APIs:

```
vm config schedule [host]
vm config coschedule [num]
```

TODO:
- [x] Add `nsmod` to change load calculation
- [x] Add documentation
- [x] Refactor after merging #815 
- [x] Test with existing models

Updates #536.


jcrussell:
Tried playing with this today, needs some tweaking before we merge.

jcrussell:
I think we should probably merge #815 first since it will clean up quite a bit of the scheduler code.

jcrussell:
This scheduler won't abort if all the available machines exceed a specified commit load. We should probably allow users to configure a limit and abort if all the machines exceed it. Do we want to do that in this PR or add it afterwards?

@djfritz

djfritz:
We should certainly consider a healthy abort condition in this PR - probably time based. This probably needs to be a tunable as well. A simple watchdog should suffice - if no progress is made in X seconds, abort.

jcrussell:
I guess we could do something time based. While the scheduler beams out VMs, it could check how long machines are taking to respond that they launched the VM and abort if that takes too long. The experiment could still blow up when you call `vm start`.

Alternatively, the scheduler is static so couldn't the abort be static too? We keep track of the VCPU/memory/network commit per-machine as we schedule VMs -- if all the machine exceed Nx commit, abort. We could support per-resource commit limits (e.g. CPU is okay at 8x, memory is okay at 4x).

jcrussell:
@djfritz -- punting on the abort condition, will address later.

PR #746, opened 2016-11-07
minimega: add locking in vncClear, other tweak

jcrussell:
Added missing locking in vncClear. vncRecordingLock > vncPlayingLock.
Return standard error rather than custom error when vm not found.


cdshann:
LGTM

PR #748, opened 2016-11-08
minimega: Small fix of function description, 'returns the VLAN for a given alias'

jasoncontrib:


djfritz:
LGTM, can you add your name to the CONTRIBUTORS file here as well?

PR #763, opened 2016-11-16
minimega: multiple dns entries for a single ip with minirouter

floren:
Previously, minimega would only assign one DNS hostname per IP, ignoring all but the
last. minirouter supported more, so a simple modification to minimega makes it work
right.


floren:
@jcrussell @djfritz

jcrussell:
Do we care about stopping users from adding the same (IP, Hostname) pair more than once?

If not, LGTM.

PR #764, opened 2016-11-17
minirouter: handle v6 static routes better

floren:
Bird was choking if you defined static routes, because it used only one
config file. If you defined ipv4 static routes, bird6 would die because
it couldn't parse the v4 routes, and vice versa. This creates /etc/bird6.conf
and makes bird6 use that instead, which has solved the problem in testing.

### DIFF
--- a/src/minirouter/bird.go
+++ b/src/minirouter/bird.go
@@ -13,11 +13,13 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


floren:
Yep, bird and bird6 both came up in my test environment.

PR #766, opened 2016-11-18
minimega/minicli: elevate preprocessor

### LINKED ISSUE #760: minimega: env preprocessor not invoked on minicli builtins
```
minimega$ echo $HOME
/root
minimega$ .csv true echo $HOME
$HOME
```

I don't remember doing that intentionally.

jcrussell:
Elevate preprocessor to minicli so that it can apply to builtins:

        minimega$ echo $HOME
        /root
        minimega$ .record false echo $HOME
        /root

Fixes #760. Need to test that this doesn't break anything.

### DIFF
--- a/src/minicli/builtins.go
+++ b/src/minicli/builtins.go
@@ -251,17 +251,6 @@ func init() {
 	}
 }
 
-func runSubCommand(c *Command) <-chan Responses {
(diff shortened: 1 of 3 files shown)


jcrussell:
Tested with #770 -- looks like it's working properly.

Issue #768, opened 2016-11-22
minimega: cliPreprocessor not run on locally launched VMs in namespace environment


jcrussell:
We don't expand `file:...` in `vm config disk` if we're running in a namespace since it should be run where the VM is launched, not configured. We expand `file:` when we launch a VM via meshage but not if it is launched locally.

@glattercj

PR #770, opened 2016-11-28
minimega: fix unpreprocessed files

jcrussell:
Add a new function to the QueuedVMs to fetch all the files that are in
the f_iombase directory which is invoked before the VMs are launched.
This applies to VMs that come in over meshage or are launched locally,
with and without a namespace. Improved the iomHelper to be smarter about
absolute paths and to supress in-flight errors.

Updates #768.


jcrussell:
Tested with #766 -- looks like it's working properly.

PR #774, opened 2016-11-29
minimega/web: Closes #772. Set no-cache headers.

### LINKED ISSUE #772: web: set no-cache on /connect/ URLs
Prevent caching of the container console / KVM VNC page.

jasoncontrib:


jcrussell:
Should we do this everywhere in web? All the content is dynamically generated.

jasoncontrib:
I don't think it's necessary to put it everywhere in web.

/connect/ can change what HTML it serves during a single run of minimega (since you might launch a container VM, destroy it, and then a KVM VM with the same name), so it definitely needs the no-cache headers.

The other URLs are not really dynamic, they serve the same HTML each time, unless minimega upgrades happen which change the HTML being served. minimega upgrades might indeed be a case we're concerned about, but it should be very infrequent.

The JSON endpoints are truly dynamic, but the AJAX GET request on the browser-side is (should be) avoiding requesting cached versions of the JSON.

jcrussell:
Sounds good.

PR #781, opened 2016-11-30
build: refactor build/test scripts, add exit codes, and add tests for minitest, go vet, and go fmt. configure for Travis CI.

jasoncontrib:


jasoncontrib:
@floren @jcrussell @djfritz 

I believe John has received an email to allows access to the minimega organization from Travis CI?

jasoncontrib:
Example Travis CI output:

https://travis-ci.org/jasoncontrib/minimega/builds/180211445

jcrussell:
Pretty neat!

jasoncontrib:
Fixed some tiny gofmt and go vet issues, and removed the python bindings build and test. Passes build on travis cleanly now:

https://travis-ci.org/jasoncontrib/minimega/builds/180487648

lint.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
jcrussell: Let's omit this for now -- it's too noisy.

fix.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
jcrussell: Let's omit this for now.

all.bash
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+bash checkformat.bash
+bash vet.bash
```
jcrussell: Maybe combine `checkformat.bash` and `vet.bash` into `check.bash`? Probably should be invoked after build.bash.

addversion.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
jcrussell: I'd rather duplicate this code in `build.bash` and `test.bash` than have another script in the directory.

build.bash
```diff
@@ -37,14 +28,27 @@ for i in `ls $SCRIPT_DIR/src | grep -v vendor`
 do
 	echo $i
 	go install $i
+	if [[ $? != 0 ]]; then
+	    LINUX_PACKAGES_ERRORS=true
```
jcrussell: Why not `exit 1` here?

jasoncontrib: Intention was to continue building the rest of the packages anyway to see if there are any other issues.

.travis.yml
```diff
@@ -0,0 +1,26 @@
+dist: trusty
+sudo: required
+language: go
+go:
+  - 1.7.3
+# for container-based testing:
+# addons:
+#   apt:
```
jasoncontrib: I'll go ahead and try to switch it over to container-based builds in Travis CI.

build.bash
jcrussell: When tweaking a dependency, we can break multiple tools. It seems better to abort the build on the first error (even though this is not the existing behavior).

Thoughts? @djfritz

.travis.yml
```diff
@@ -1,26 +1,17 @@
...
-#     - openvswitch-switch # not in travis-ci whitelist yet
-before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y libpcap-dev libreadline-dev qemu qemu-kvm openvswitch-switch dnsmasq python-jinja2
+addons:
+  apt:
+    packages:
+    - libpcap-dev
+    - libreadline-dev
+    - python-jinja2
```
jcrussell: Can remove.

jcrussell:
Are we ready to merge this?

jasoncontrib:
I think @djfritz wants to use our own build box rather than Travis. In that case, we can run minitests, too. Should I add back the minitest.bash script and get rid of the Travis config file?

djfritz:
Right. Shouldn't we want to have the full scope of tests we have all occur in our test harness? If so, then we should just start with internal tests that maybe can be exported publicly.

jcrussell:
You can have multiple integrations, right? Why not use travis for all that it can do and then we can figure out minitests later?

jcrussell:
[0] shows that Github support for multiple CI services -- I say we merge this and then add a second `minitest`-based CI service later.

[0] https://github.com/blog/1935-see-results-from-all-pull-request-status-checks

djfritz:
In that case, sounds good.

jcrussell:
LGTM.

What do we need to do to enable Travis integration with the main repo?

djfritz:
There's a pending request. I'll enable it before we merge.

jcrussell:
Sounds good, enable and merge away when you can.

PR #785, opened 2016-12-01
minimega: prepare to deprecate targeting VMs by ID. Updates #680.

### LINKED ISSUE #680: minimega: deprecate targeting VMs by ID
Namespaces makes targeting VMs by ID ambiguous so we should eliminate the ability to target VMs by ID.

jasoncontrib:


src/minimega/vm_cli.go
```diff
@@ -303,7 +295,7 @@ name, and a JSON string, and returns the JSON encoded response. For example:
 		HelpLong: `
 Take a screenshot of the framebuffer of a running VM. The screenshot is saved
 in PNG format as "screenshot.png" in the VM's runtime directory (by default
-/tmp/minimega/<vm id>/screenshot.png).
+/tmp/minimega/<vm name>/screenshot.png).
```
jcrussell: Hmm. We probably will need to do something like:

`<namespace>/<name>`

src/minimega/vmlist.go
```diff
@@ -211,12 +211,7 @@ func (vms VMs) FindVMNoNamespace(s string) VM {
 // findVM assumes vmLock is held.
 func (vms VMs) findVM(s string, checkNamespace bool) VM {
 	if id, err := strconv.Atoi(s); err == nil {
-		if vm, ok := vms[id]; ok {
-			if inNamespace(vm) || !checkNamespace {
-				return vm
-			}
-		}
-
+		log.Warn("Targeting VMs by ID has been deprecated! (VM %d)", id)
```
jcrussell: Not sure we want to go this far yet... Need to think about this.

src/minimega/vmlist.go
```diff
@@ -563,10 +557,12 @@ func (vms VMs) apply(target string, concurrent bool, fn vmApplyFunc) []error {
 	for _, v := range vals {
 		id, err := strconv.Atoi(v)
 		if err == nil {
-			ids[id] = true
-		} else {
-			names[v] = true
+			log.Warn("Targeting VMs by ID has been deprecated! (VM %d)", id)
```
jcrussell: See above.

src/minimega/vm_cli.go
jasoncontrib: A few questions about paths:
* What path to use if there is no active namespace? `/tmp/minimega/<name>/...`?
And if so, what to do about a host without a namespace that has the same name as an active namespace? E.g. collision between `/tmp/minimega/<namespace>/...` and `/tmp/minimega/<name>/...` where `<namespace>==<name>`
* Other VM files are in `/tmp/minimega/<vm id>/...`, do we now break this convention for the screenshots?

src/minimega/main.go
```diff
@@ -50,6 +50,7 @@ var (
 	f_cli        = flag.Bool("cli", false, "validate and print the minimega cli, in JSON, to stdout and exit")
 	f_panic      = flag.Bool("panic", false, "panic on quit, producing stack traces for debugging")
 	f_cgroup     = flag.String("cgroup", "/sys/fs/cgroup", "path to cgroup mount")
+	f_future     = flag.Bool("test-the-future", false, "disable features that will be removed in future releases, such as targeting VMs by ID")
```
jcrussell: Hm, I was thinking the opposite direction for this flag -- log a warning if they target VMs by ID and don't specify `-i-wont-index-vms-by-id-in-the-future` (or something similar).

jasoncontrib: It does log a warning if they target VMs by ID and don't specify the flag.

The flag disallows targeting VMs by ID and does not log a warning, so it achieves the same behavior as the future release in which targeting VMs by ID is disabled.

Or did you mean there should be a flag purely to disable the warnings, and still allow targeting VMs by ID?

jasoncontrib: Added a `-i-wont-index-vms-by-id-in-the-future` flag with that behavior.

jcrussell: On further discussion, remove the flag and the warning -- we'll make those changes in 2.4.

PR #794, opened 2016-12-08
minimega: use ioctl and partx to rescan devices

jcrussell:
Works ~99% of the time! Good find Jason.

Supersedes #791.


jasoncontrib:
Works ~25% of the time on my system :(

jasoncontrib:
#801 should work ~100% of the time...

Issue #731, opened 2016-10-26
minimega: minicli Header uniformity


jcrussell:
It would be nice if all the Headers were lowercase and a single word. This annoys me:

`.columns Bridge,"Trunk ports"  bridge`

jcrussell:
We should also move units to the column headers rather than have the units in every row.

jasoncontrib:
Is there a way to add units to the column headers without having to type them when using the .columns filter syntax?

jcrussell:
We support prefixes elsewhere -- we could do with `.columns` as well.

djfritz:
Are you talking about the apropos support we have in minicli?

jcrussell:
Yup. What do you think?

djfritz:
Sounds perfect to me - the uninformed user would probably expect to be able to shorthand column names just the same as the cli terminals anyway.

jasoncontrib:
Do we want to support apropos matches / prefixes with .filter, too?

jcrussell:
Yup.

jcrussell:
Closed by #800

PR #800, opened 2016-12-12
minicli: Updates #731. Start addressing minicli Header uniformity.

jasoncontrib:
Opening a PR now so we can start discussing specifics. Need to fix minitests to match once we've decided on column names (and units?).


jasoncontrib:
Good point, for some reason I was (mistakenly) worried about not appending to the tabular unless the function was going to return cleanly. Just needs a `columns[i] = header` so that the returned header uses the full column names and not the incomplete column names.

jasoncontrib:
We may have to have some ugliness in the code (or an extra loop) after all, to avoid the corner case where an exact match is found AFTER a partial match, e.g.

```
minimega$ .filter name~host .col namespace,name,st vm info
Error (tinkerpad): ambiguous column `name`
minimega$ .filter name~host .col name,namespace,st vm info
host      | name  | namespace | state
tinkerpad | host1 |           | RUNNING
tinkerpad | host2 |           | RUNNING
```

src/minicli/builtins.go
```diff
@@ -280,6 +281,34 @@ func parseFilter(s string) (filter, error) {
...
+		if foundI >= 0 {
+			// collision
+			return 0, fmt.Errorf("ambiguous column `%s`", column)
+		}
+
+		foundI = i
+	}
+
+	if foundI >= 0 {
+		return foundI, nil //TODO
```
jcrussell: What is TODO for?

jasoncontrib: Oops, forgot to remove the TODO. It does work and passes the `builtins` minitest with apropos / partial matching additions.

PR #801, opened 2016-12-12
minimega: fix disk inject sometimes not working

jasoncontrib:


src/minimega/disk.go
```diff
@@ -155,34 +161,51 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
...
-	if err != nil {
-		return err
-	}
+	path := nbdPath
 
-	syscall.Syscall(syscall.SYS_IOCTL, f.Fd(), C.BLKRRPART, 0)
+	// decide whether to mount partition or raw disk
+	if partition != "none" {
+		// default to first partition if there is only one partition
+		if partition == "" {
```
jcrussell: Move this to after we have detected that there are partitions on the device.

src/minimega/disk.go
```diff
@@ -155,34 +161,51 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
...
-		}
+		path = nbdPath + "p" + partition
 
-		partition = "1"
-	}
+		// keep rereading partitions and waiting for them to show up for a bit
+		timeoutTime := time.Now().Add(1 * time.Second)
+		for {
+			// tell kernel to reread partitions
+			f, err := os.Open(nbdPath)
```
jcrussell: Needs closing. Possibly open once and then syscall multiple times.

src/minimega/disk.go
```diff
@@ -155,34 +161,51 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
...
+
+			time.Sleep(100 * time.Millisecond)
+		}
+
+		// check desired partition exists
+		_, err = os.Stat(path)
+		if err != nil {
+			return errors.New("desired partition not found on image!")
+		} else {
+			log.Debug("desired partition found on image")
```
jcrussell: Maybe add the partition number/disk here.

src/minimega/disk.go
```diff
@@ -155,34 +161,51 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
...
-	if partition == "none" {
-		path = nbdPath
-	} else {
-		path = nbdPath + "p" + partition
+			_, err = os.Stat(nbdPath + "p1")
+			if err == nil {
+				break
+			}
+
+			if time.Now().After(timeoutTime) {
```
jcrussell: Maybe make loop condition.

tests/disk.want
```diff
@@ -9,18 +9,19 @@ E: Information: You may need to update /etc/fstab.
 
 
 ## shell mkfs.fat /dev/nbd0p1
-mkfs.fat 3.0.28 (2015-05-16)
-unable to get drive geometry, using default 255/63
+mkfs.fat 4.0 (2016-05-06)
```
jcrussell: Should we suppress output here too?

jcrussell:
Worked 100/100 times for me -- LGTM.

PR #782, opened 2016-12-01
minilog and friends: elevate logSetup, refactor minilog importers

jasoncontrib:

PR #817, opened 2016-12-23
protonuke/vmbetter/vncdrone: use log.Init

jcrussell:
Somehow these were missed in #782. Caused a panic when running vmbetter
since the flags were already defined.


jasoncontrib:
LGTM

PR #818, opened 2016-12-24
[rfc] minimega: `vm top` API

jcrussell:
Example output:

```
name | namespace | virt (MB) | res (MB) | shr (MB) | cpu (%) | vcpu (%) | time+
vm-0 |           | 2751      | 1382     | 3        | 0.00    | 0.00     | 6s
vm-1 |           | 2740      | 1382     | 3        | 0.00    | 0.00     | 6s
```

Pulls qemu process utilization from proc. Mostly works but the CPU calculation seems a bit off. Should be possible to walk the process tree for containers too but that is not implemented yet.

Should roll the bandwidth stats into this API rather than `vm info`.

We could also pull some IO stats for disk activity.

Thoughts?


jcrussell:
`watch bin/minimega -e vm top` looks pretty slick.

djfritz:
Screenshot or it didn't happen.

jcrussell:
Video or it didn't happen?

PR #819, opened 2016-12-30
minimega/minicli/miniclient: support completion over -attach

### LINKED ISSUE #600: minimega: completion on `attach`
If you try to tab complete over the `attach` interface, it may not complete properly or, in some cases, panic. We should make sure that it doesn't panic, at least. If we can, it would be great to have tab completion work the same as the regular CLI.

Known panics:
- Everything that uses `cliVMSuggest`
  - Caused by `meshageNode` not being initialized

Perhaps we need a way to send minimega a partial string to complete. Right now the interface between miniclient and minimega is minicli.Command <-> miniclient.Response. We could mux more things through the channel as done by meshage.

jcrussell:
Send suggestions over the -attach connection. Fixes #600.

### DIFF
--- a/src/minicli/handler.go
+++ b/src/minicli/handler.go
@@ -28,9 +28,9 @@ type Handler struct {
 	// Suggest provides suggestions for variable completion. For example, the
 	// `vm stop` command might provide a listing of the currently running VM
 	// names if the user tries to tab complete the "target". The function takes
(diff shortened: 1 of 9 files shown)


djfritz:
@jcrussell is this ready to merge?

jcrussell:
Yup.

PR #821, opened 2016-12-30
minimega/nbd/bridge: remove process aliases

### LINKED ISSUE #787: minimega: remove `defaultExternalProcesses`
What is the purpose of this map? It provides aliases for commands but cannot be updated. We do use it to check for all runtime dependencies which is useful. I suggest we remove it to remove the indirection.

jcrussell:
mininega: `defaultExternalProcesses` provided a mapping from our alias for a process to the actual binary name. Removed the aliases and instead enumerate the names of all the processes. Documented where external dependencies are used and added dependencies for imported packages.

bridge/nbd: enumerate all the external dependencies for the respective packages in an exported variables.

Fixes #787


src/minimega/external.go
```diff
@@ -26,40 +26,37 @@ var (
...
-	"taskset":  "taskset",
-	"lsmod":    "lsmod",
-	"ntfs-3g":  "ntfs-3g",
-	"scp":      "scp",
-	"ssh":      "ssh",
-	"hostname": "hostname",
-	"tc":       "tc",
+// externalDependencies contains all the external programs that minimega
+// invokes. We check for the existence of these on startup and on `check`.
+var externalDependencies = map[string]bool{
```
djfritz: Should we remove the dependencies on kill, mount (though I think there's a special need for the mount binary somewhere), hostname, cp, umount, and rm?

jcrussell: We definitely need `mount` for something -- it does some magic that's not replicated with the `syscall.Mount`.

I can replace the other dependencies with stdlib calls.

jcrussell: We could replace `tastset` too using `sched_setaffinity`.

jcrussell:
Trimmed a lot of the dependencies. I don't really want to implement `cp -rf` or `taskset` in this PR.

PR #842, opened 2017-01-19
minimega: add container quotas

### LINKED ISSUE #837: minimega: add container CPU quotas
We could use `cpu.cfs_period_us` and `cpu.cfs_quota_us` to limit tasks based on `vm config vcpus`.

Specifically:
```
cpu.cfs_period_us = 1000000 // 1 second
cpu.cfs_quota_us = $VCPUs * cpu.cfs_period_us
```

Where `$VCPUs` is the `vm config` value.

jcrussell:
Enforces the VCPUs set via `vm config vcpus`.

Fixes #837.

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -298,7 +298,8 @@ func containerInit() error {
 	cgroupFreezer := filepath.Join(*f_cgroup, "freezer", "minimega")
 	cgroupMemory := filepath.Join(*f_cgroup, "memory", "minimega")
 	cgroupDevices := filepath.Join(*f_cgroup, "devices", "minimega")
-	cgroups := []string{cgroupFreezer, cgroupMemory, cgroupDevices}
+	cgroupCPU := filepath.Join(*f_cgroup, "cpu", "minimega")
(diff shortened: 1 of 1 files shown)


src/minimega/container.go
```diff
@@ -365,15 +367,16 @@ func containerTeardown() {
...
+//	6 :  memory in megabytes
+//	7 :  uuid
+//	8 :  number of fifos
+//	9 :  preinit program
+//	10 :  init program (relative to filesystem path)
+//	11:  init args
 func containerShim() {
 	args := flag.Args()
-	if flag.NArg() < 10 { // 10 because init args can be nil
+	if flag.NArg() < 11 { // 10 because init args can be nil
```
djfritz: Fix the comment

src/minimega/container.go
```diff
@@ -1359,9 +1367,21 @@ func containerPopulateCgroups(vmID, vmMemory int) error {
 		}
 	}
 
+	// vcpus
```
djfritz: can we reference something that indicates that the period/quota calculation here maps to an integer cpu unit?

PR #844, opened 2017-01-19
minimega: add time in state to vm info and web interface

### LINKED ISSUE #811: minimega: uptime per each VM
Create a per-VM uptime record.  May exist but not shown on web.

jasoncontrib:
Closes #811.

### DIFF
--- a/src/minimega/vm.go
+++ b/src/minimega/vm.go
@@ -45,7 +45,7 @@ type VM interface {
 	GetNetworks() []NetConfig // GetNetworks returns an ordered, deep copy of the NetConfigs associated with the vm.
 	GetHost() string          // GetHost returns the hostname that the VM is running on
 	GetState() VMState
-	GetStateStart() time.Time // GetStateStart returns the time when the VM entered the current state
+	GetLaunchTime() time.Time // GetLaunchTime returns the time when the VM entered the current state
 	GetType() VMType
 	GetInstancePath() string
 	GetUUID() string
@@ -109,7 +109,7 @@ type BaseVM struct {
 	Host      string // hostname where this VM is running
 
 	State      VMState
-	StateStart time.Time
+	LaunchTime time.Time
 	Type       VMType
 	ActiveCC   bool // set when CC is active
 
@@ -207,7 +207,7 @@ func NewBaseVM(name, namespace string, config VMConfig) *BaseVM {
 	vm.instancePath = filepath.Join(*f_base, strconv.Itoa(vm.ID))
 
 	vm.State = VM_BUILDING
-	vm.StateStart = time.Now()
+	vm.LaunchTime = time.Now()
 
 	// New VMs are returned pre-locked. This ensures that the first operation
 	// called on a new VM is Launch.
@@ -227,7 +227,7 @@ func (vm *BaseVM) copy() *BaseVM {
 	vm2.Namespace = vm.Namespace
 	vm2.Host = vm.Host
 	vm2.State = vm.State
-	vm2.StateStart = vm.StateStart
+	vm2.LaunchTime = vm.LaunchTime
 	vm2.Type = vm.Type
 	vm2.ActiveCC = vm.ActiveCC
 	vm2.instancePath = vm.instancePath
@@ -307,11 +307,11 @@ func (vm *BaseVM) GetState() VMState {
 	return vm.State
 }
 
-func (vm *BaseVM) GetStateStart() time.Time {
+func (vm *BaseVM) GetLaunchTime() time.Time {
 	vm.lock.Lock()
 	defer vm.lock.Unlock()
 
-	return vm.StateStart
+	return vm.LaunchTime
 }
 
 func (vm *BaseVM) GetType() VMType {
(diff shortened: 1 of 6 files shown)


src/minimega/vm.go
```diff
@@ -620,8 +635,10 @@ func (vm *BaseVM) Info(field string) (string, error) {
 // setState updates the vm state, and write the state to file. Assumes that the
 // caller has locked the vm.
 func (vm *BaseVM) setState(s VMState) {
-	log.Debug("updating vm %v state: %v -> %v", vm.ID, vm.State, s)
+	now := time.Now()
+	log.Debug("updating vm %v state: %v -> %v, at time %s", vm.ID, vm.State, s, now)
```
jcrussell: `log` includes the current time -- it doesn't seem necessary to add it to the message.

src/minimega/vm.go
```diff
@@ -568,6 +581,8 @@ func (vm *BaseVM) Info(field string) (string, error) {
 		return vm.Namespace, nil
 	case "state":
 		return vm.State.String(), nil
+	case "statetime":
```
jcrussell: Maybe `updated` makes more sense that `statetime` and instead of `time.Since`, report the actual time that the VM was mutated?

src/minimega/vm.go
```diff
@@ -44,6 +45,7 @@ type VM interface {
 	GetNetworks() []NetConfig // GetNetworks returns an ordered, deep copy of the NetConfigs associated with the vm.
 	GetHost() string          // GetHost returns the hostname that the VM is running on
 	GetState() VMState
+	GetStateStart() time.Time // GetStateStart returns the time when the VM entered the current state
```
jcrussell: Glad `miniweb` will allow us to drop a lot of these accessor methods.

src/minimega/vm.go
jasoncontrib: Used `time.Since` since @bjwrigh asked for something like `uptime`, which reports a duration?

djfritz:
I read the discussion in #811 as just the total wallclock uptime from `vm launch`. What's the use case to reset the time at each state change?

jasoncontrib:
I read @bjwrigh 's original request for `uptime` as wanting to know wallclock time that a VM had been `RUNNING`, and figured we might as well show it for other states than `RUNNING`. Perhaps we can get some clarification from @bjwrigh ?

bjwrigh:
Only the RUNNING state seems necessary.  Clearing the time when the VM leaves the RUNNING state would probably help keep things accurate.  I don't see an added benefit to seeing how long something was BUILDING, other than maybe using that as a trigger to do a distributed start on the VM's for resource purposes - but that information can be discerned from the level of the script running the experiment.

tl;dr:  My opinion is to only track RUNNING time, reset to n/a or zero when not in that state.

jasoncontrib:
Would you want to see time as duration elapsed, or the date/time at which it started in the RUNNING state?

________________________________

bjwrigh:
I don't know that I have a way of choosing which one would be best in any way other than preference.  The elapsed time may be deceiving on a stale info page, while the start time might be deceiving if no date is present, or the system clock on one node of the cluster is off.  Both would lead to it being too cluttered.  I'm leaning towards uptime... @djfritz thoughts?

jcrussell:
It seems like elapsed time makes the most sense if we can't count on system clocks being correct/synchronized.

@jasoncontrib: I would tweak this to be called `uptime` and only apply to VMs in the running state.

However, if a VM is paused/started, should the uptime revert to 0?

djfritz:
I still don't see the point. If I pause and later unpause a QEMU vm, the vm will see the jump in time. If the uptime is just time in the running state, what possible use is there?

My use case for something like this is to have a simple timestamp of when the vm was launched, rendered in vm info as an elapsed wall clock time from that launch point. I'd use that as an indicator of things like when I launch a new experiment and find some old VM lingering and messing up my experiment.

src/minimega/vm.go
```diff
@@ -302,6 +307,14 @@ func (vm *BaseVM) GetState() VMState {
 	return vm.State
 }
 
+func (vm *BaseVM) GetStateStart() time.Time {
+	vm.lock.Lock()
+	defer vm.lock.Unlock()
+
+	log.Info("Returning StateStart: ", vm.StateStart.String())
```
jcrussell: Don't think this log is necessary.

jcrussell:
@djfritz: If that's the case, then we should store `Started` and always compute `time.Since(vm.Started)`?

src/minimega/vm.go
```diff
@@ -44,6 +45,7 @@ type VM interface {
 	GetNetworks() []NetConfig // GetNetworks returns an ordered, deep copy of the NetConfigs associated with the vm.
 	GetHost() string          // GetHost returns the hostname that the VM is running on
 	GetState() VMState
+	GetLaunchTime() time.Time // GetLaunchTime returns the time when the VM entered the current state
```
jcrussell: Forgot to update comment.

PR #845, opened 2017-01-20
bridge/minimega: replace gopcap with gopacket

### LINKED ISSUE #331: minimega: rewrite capture api to use gopacket
see summary. The current implementation is bad.

jcrussell:
Remove gopcap and use gopacket instead. Move the capture logic into
bridge. Added support for BPF filters to capture, could add that to the
CLI. You can also now do multiple captures per-bridge or per-tap with
different filters. Needs more testing but initial results show it's
pretty stable.

Fixes #331.


djfritz:
Think this is still in the 2.4 window?

jcrussell:
I haven't tested this on debian testing which has a newer version of libpcap. On my machine, it is more stable than the old code. We could probably test and merge this for 2.4.

FWIW, we could add support for BPF to the capture API -- it's supported in the bridge package now.

jcrussell:
Do we want to add BPF support to the API? I would also find it useful to configure the snaplen (in cases where I only want headers).

New APIs:

```
capture filter [filter]
capture snaplen [snaplen]
```
these would apply to *new* captures, similar to `vm config`.

djfritz:
BPF strings would be nice

floren:
Jon, what's the status on this? We've actually had another reason come up to want to use native Go 100% instead of using C bindings, and this is a major step (getting rid of readline is the other)

jcrussell:
I've been using some parts of this and it's more stable than the existing capture code. I need to figure out what to do with the "netflow timeouts" as that is the only thing that isn't namespaced. We also need to do some testing with newer versions of libpcap.

AFAIK, there isn't a native Go libpcap -- so minimega would build but the capture API wouldn't work, right?

jcrussell:
From the OVS database schema documentation [0], it looks like we cannot have different "netflow timeouts" per namespace -- there can be only one netflow object per bridge. Using `capture` on a bridge is inherently non-namespacey since all VM traffic passes through the same bridge (usually). You can still, however, have multiple netflow writers (these are tracked per-namespace). I'm thinking that we should just mark the following APIs as "use with caution with namespaces" and move on:

```
        capture netflow bridge <bridge> <filename>
        capture netflow bridge <bridge> <tcp,udp> <hostname:port>
        capture netflow timeout [timeout in seconds]
        capture pcap bridge <bridge> <filename>
```

The following APIs should work as expected with namespaces:

```
        capture
        capture pcap snaplen [size]
        capture pcap filter [bpf]
        capture netflow mode [raw,ascii]
        capture netflow gzip [true,false]
        capture pcap vm <name> <interface index> <filename>
        capture pcap delete vm <name>
        capture netflow delete bridge <name>
        capture pcap delete bridge <name>
```

Thoughts? @djfritz @floren 

[0] http://openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt

jcrussell:
Another option: we could implement our own netflow tool using gopacket.

jcrussell:
```
for i in $(seq 1000); do bin/minitest -run capture -level info; done
```

Finished and then I cleanly `quit` minimega.

PR #849, opened 2017-02-02
minilog/ron/miniccc/minimega: refactoring, add update of miniccc level at runtime

### LINKED ISSUE #705: minimega/ron/miniccc: change log level from cc
It would be convenient to be able to change `miniccc`'s log level from a `cc` API.

jcrussell:
Some refactoring in minilog: created a type for Level.

Added new minimega API: `cc log level ...` to update miniccc's log level at runtime.

Fixes #705


jcrussell:
Adding a type for iota is part of the stringer example:

https://godoc.org/golang.org/x/tools/cmd/stringer

PR #852, opened 2017-02-03
miniweb: is born

jcrussell:
_Don't merge yet, still a WIP._

Somewhat functional miniweb. vms, tilevnc, and connect all seem to be
working. Need to fix the graph view.


jcrussell:
`miniweb` is now functionally the same as minimega's web server. Do we want to do a review and merge this and then figure out what things we want to add?

@djfritz @jasoncontrib @floren

jcrussell:
It would be neat to embed all of "misc/web" into the miniweb binary. Then we would have a single binary web. Something like this looks promising:

https://github.com/rakyll/statik

Should we create a bunch of miniweb issues for all the different features we want?

floren:
sure

jasoncontrib:
LGTM, and I've got the embedding of the `misc/web` files working using `statik`

djfritz:
Do we really care to bundle the web content in the binary these days? Anymore, we tend to package the minimega distribution as a whole - there are already tons of binaries, docs, etc.

jcrussell:
What's the verdict on static `miniweb`? Are we moving towards a single directory that needs to be replicated on the head node? `minidoc` could embed all of `doc/`...

PR #853, opened 2017-02-03
minimega: vince is unstoppable

jcrussell:
VMs that share his name inherit his unstoppable will.


activeshadow:
Yeah... this won't cause any bugs in the future. :smile:

PR #855, opened 2017-02-03
Protonuke: Add FTP support

csymonds:
Adding in FTP traffic generation to protonuke
Supports TLS via -ftps

Client will connect and authenticate to a server, randomly choosing one of 6 ftp actions (including logging out) with uniform distribution.

File retrieval (RETR) is accomplished by sending an image file of specified size (default 500KB) over a passive data connection. The image is generated in the same way as http.go. 

2 vendor packages were added to handle server and client implementation (with some additional implementation and updates required)


jcrussell:
Woah... I think something about the `git rebase` went wrong. There shouldn't be other people's commits here.

csymonds:
The rebase resulted in a series of conflicts which in turn resulted in additional commits. After deliberating with Jon, we have decided to add a versions file to vendor for tracking package information instead of a rebase. I will be deleting this PR and creating a new one with a new branch.

PR #859, opened 2017-02-08
noVNC - Now with more buttons

mkunz7:
Simplifies code.

Enables scaling, mobile keyboard, mobile l/m/r click, viewport, and more.
Clipboard is hidden because QEMU sockets do not provide functionality.

novnc


x64x6a:
doesn't work for me
novnc

mkunz7:
clear cache

x64x6a:
It work's if I open it in incognito..
novnc1

x64x6a:
The settings menu seems sort of useless though?  Should you hide it?
Also, the icon they have for alt+ctrl+del isn't very obvious..

mkunz7:
If you disconnect you can use it to change the things like scaling mode.

mkunz7:
The websocket for novnc doesn't close switching between machines. Closing the browser seems to be the only way. Hold off on the merge.

mkunz7:
Fixed the websocket issue, should be safe to merge.

djfritz:
@mkunz7 can you bring this up to HEAD and we'll figure out getting this in.

PR #862, opened 2017-02-13
Changed VNC page title to VM name.

glattercj:
Previously it was set to just the hostname of the server that served the html file. It makes more sense for it to be the name of the VM.


djfritz:
Update this to include your name in the CONTRIBUTORS file.

Issue #879, opened 2017-02-27
igor: ssh key / root password set at reservation time


floren:
When igor creates a reservation, the user should be able to specify either an SSH key or a password to be installed for root. igor should unpack the initrd and install the key or password.

floren:
#885 fulfills this.

PR #885, opened 2017-03-01
Passwordify

floren:
Add passwordify, which lets you set a root password and (if desired) authorized_keys to an initramfs.

### DIFF
--- a/src/passwordify/external.go
+++ b/src/passwordify/external.go
@@ -11,12 +11,13 @@ import (
 )
 
 var externalProcesses = map[string]string{
-	"bash":   "bash",
-	"chroot": "chroot",
-	"find":   "find",
-	"cpio":   "cpio",
-	"zcat":   "zcat",
-	"gzip":   "gzip",
+	"bash":       "bash",
+	"chroot":     "chroot",
+	"find":       "find",
+	"cpio":       "cpio",
+	"zcat":       "zcat",
+	"gzip":       "gzip",
+	"ssh-keygen": "ssh-keygen",
 }
 
 // check for the presence of each of the external processes we may call,
--- a/src/passwordify/external.go
+++ b/src/passwordify/external.go
@@ -13,6 +13,10 @@ import (
 var externalProcesses = map[string]string{
 	"bash":   "bash",
 	"chroot": "chroot",
+	"find":   "find",
+	"cpio":   "cpio",
+	"zcat":   "zcat",
+	"gzip":   "gzip",
 }
 
 // check for the presence of each of the external processes we may call,
--- a/src/passwordify/external.go
+++ b/src/passwordify/external.go
@@ -11,8 +11,8 @@ import (
 )
 
 var externalProcesses = map[string]string{
-	"bash":        "bash",
-	"chroot":	"chroot",
+	"bash":   "bash",
+	"chroot": "chroot",
 }
 
 // check for the presence of each of the external processes we may call,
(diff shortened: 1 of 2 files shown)


src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+	// Set password
+	p := process("chroot")
+	cmd := &exec.Cmd{
+		Path: p,
+		Args: []string{
+			p,
+			tdir,
+			"passwd",
+		},
+		Env: nil,
```
jcrussell: Can leave these off if using zero value.

src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+			log.Fatalln("passwords do not match")
+		}
+		io.WriteString(stdin, string(pw1)+"\n")
+		io.WriteString(stdin, string(pw1)+"\n")
+	}()
+
+	log.LogAll(stdout, log.INFO, "chroot")
+	log.LogAll(stderr, log.INFO, "chroot")
+
+	cmd.Run()
```
jcrussell: Would be a bit cleaner to use cmd.CombinedOutput() and ditch stdout/stderr pipes.

src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	go func() {
+		defer stdin.Close()
+		fmt.Printf("Enter new root password: ")
```
jcrussell: Could put this in a for loop so we keep reading until we get two passwords that match (or an error).

src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// If keyfile, copy keyfile
+	if *f_keys != "" {
+		in, err := os.Open(*f_keys)
+		if err != nil {
+			log.Fatalln("can't open key file source:", err)
+		}
+		defer in.Close()
+
+		err = os.Mkdir(tdir+"/root/.ssh", os.ModeDir|0700)
```
jcrussell: Use `filepath.Join`?

src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			log.Fatalln("can't make root's ssh directory:", err)
+		}
+
+		out, err := os.OpenFile(tdir+"/root/.ssh/authorized_keys", os.O_RDWR|os.O_CREATE, 0600)
```
jcrussell: Use `filepath.Join`?

src/passwordify/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	// Cleanup
+	err = os.RemoveAll(tdir)
+	if err != nil {
+		log.Fatalln(err)
+	}
+}
+
+func runscript(cmdString string) error {
```
jcrussell: It'd be simpler to use `bash -c "..."` instead, this seems like it would work though.

src/passwordify/external.go
```diff
@@ -0,0 +1,39 @@
...
+
+package main
+
+import (
+	"fmt"
+	log "minilog"
+	"os/exec"
+)
+
+var externalProcesses = map[string]string{
```
jcrussell: Add `cd`, `find`, `cpio`, `bash`, and `zcat`?

src/passwordify/external.go
```diff
@@ -0,0 +1,39 @@
...
+	"bash":   "bash",
+	"chroot": "chroot",
+}
+
+// check for the presence of each of the external processes we may call,
+// and error if any aren't in our path
+func externalCheck() {
```
jcrussell: Call this in `main`?

floren:
PTAL

floren:
btw this is supposed to provide #879

floren:
Forgot that I wanted to allow nodes booted with an image to ssh to each other using ssh keys / no password, @djfritz @jcrussell can you check again?

PR #886, opened 2017-03-02
igor: fix concurrent map write, faster runShow

jcrussell:
Fix concurrent map write to nodes. Concat'ing strings is slow so use
bytes.Buffer. Move findReservation to preprocessing step.


floren:
LGTM, this has been neglected for a long time and those changes make sense.

PR #887, opened 2017-03-06
misc: add cpu,cpuacct to cgroups

jcrussell:
In #842, we added support for CPU quotas which requires the cpu,cpuacct
cgroup. Adding this to our host images.


jcrussell:
Still need to test this.

djfritz:
Is this ready to merge?

jcrussell:
I've been running these commands manually to launch containers on the cluster -- should be fine to merge. Will need to update the pre-built image.

jcrussell:
@djfritz -- tested, this seems to be working completely.

PR #888, opened 2017-03-13
rond: back from the dead

jcrussell:
I've found this to be very useful for controlling machines running miniccc without minimega.

It's pretty well tested as a standalone service. I haven't really tested how the changes affect minimega (shouldn't really, only added a flag to `ron.Server`).


djfritz:
@jcrussell have a fix for ron.go so I can merge?

PR #889, opened 2017-03-13
minimega/vmconfiger: several bug fixes, fix #854

### LINKED ISSUE #854: Question: can preinit be used to inject files from host into container filesystem?
Just curious if I can use the `vm config preinit` option to inject, as an example, static network config files into a container before it starts. If so, is there a predetermined path or env variable I can use in the preinit script to point to the location of the container's root filesystem?

If not, any chance we could get a `vm config volume` option for containers? I could take a stab at implementing it...

jcrussell:
Few related things in this PR:
 * Added `vm config volume` API, fixing #854.
 * Added support for `map[string]string` to `vmconfiger`.
 * Updated `BaseConfig.Tags` to use new map support.
 * Added missing directory in cgroup `nuke`.
 * Fixed test broken in #852.


jcrussell:
Note: I left a TODO to handle spaces in the paths.

@activeshadow: want to give this a test drive?

activeshadow:
Of course! I'll try to give it a whirl this weekend. Thx!

activeshadow:
Sorry, got waylaid last week w/ work. I'm going to take a look at this right now.

activeshadow:
@jcrussell OK, so I'm seeing some errors when I try to use the new `volume` config... perhaps I'm going about it wrong.

Here's my config (`/tmp/mm-volume-test.txt`):

```
vm config memory 512
vm config filesystem /opt/minimega/sceptre_container_rootfs/

vm config net 10
vm config volume /etc/network/interfaces /tmp/kronos-interfaces
vm launch container kronos

vm config net 10
vm config volume /etc/network/interfaces /tmp/solver-interfaces
vm config volume /tmp/6-bus.pb /tmp/6-bus.pb
vm launch container solver

cc filter hostname=kronos
cc exec sceptre-kronosd --endpoint 172.16.10.10 --command start

cc filter hostname=solver
cc exec pysceptre-dummy-power-server -d -k 172.16.10.10 -pb /tmp/miniccc/files/sceptre/6-bus.pb 2>&1 > /tmp/solver.log
```

And here's the commands I ran (as root):

```
$ minimega
minimega, Copyright (2014) Sandia Corporation.
Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
the U.S. Government retains certain rights in this software.
minimega$ read /tmp/mm-volume-test.txt
2017/03/27 12:04:02 WARN vlans.go:334: Blacklisting manually specified VLAN 10
Error (minibuntu): freezer: open /sys/fs/cgroup/freezer/minimega/0/freezer.state: no such file or directory
Error (minibuntu): freezer: open /sys/fs/cgroup/freezer/minimega/1/freezer.state: no such file or directory
minimega$ .columns name,state,type,volume vm info
host      | name   | state | type      | volume
minibuntu | kronos | ERROR | container | {"/etc/network/interfaces":"/tmp/solver-interfaces","/tmp/6-bus.pb":"/tmp/6-bus.pb"}
minibuntu | solver | ERROR | container | {"/etc/network/interfaces":"/tmp/solver-interfaces","/tmp/6-bus.pb":"/tmp/6-bus.pb"}
```

Since I launched the `kronos` container prior to changing the volume for `/etc/network/interfaces` and adding the volume for `/tmp/6-bus-pb`, I would have expected the host volume to be different for each container, and I wouldn't have expected to see the `/tmp/6-bus.pb` volume on the `kronos` container. However, it seems as though the last volume setting is getting applied to both containers.

Do the source/target have to be directories, or can they be actual files too (like I'm assuming they can be here)?

jcrussell:
I know what the bug is -- need to deep copy a map. I can fix it later today.

Hm, I had only considered directories initially but it might be possible to handle files as well. @djfritz, what do you think?

activeshadow:
Glad I could help identify the bug!

Files vs directories isn't critical. I'm used to it from Docker but directories only works just as good!

jcrussell:
@activeshadow -- that bug should be fixed. You still won't be able to use files as volumes though.

activeshadow:
@jcrussell OK, I'll test it out shortly and let you know how it goes. Thanks!

djfritz:
While you can bind mount an individual file, it's dangerous. If something like vim rewrites the file (which is does by default), a new inode will get created, effectively undoing the bind mount. The only safe way to write to a bind mounted file is to append to it, which isn't terribly useful. 

How does docker do this? A symlink?

djfritz:
docker has the same issue:

https://github.com/docker/docker/issues/6011

djfritz:
From the docker docs:

Note: Many tools used to edit files including vi and sed --in-place may result in an inode change. Since Docker v1.1.0, this will produce an error such as “sed: cannot rename ./sedKdJ9Dy: Device or resource busy”. In the case where you want to edit the mounted file, it is often easiest to instead mount the parent directory.

djfritz:
So, no, I suspect we won't support individual files anymore than docker does.

djfritz:
@jcrussell maybe support this in the same way but bind mount non-directories as RO and throw a warning?

jcrussell:
It sounds like directories are good enough for @activeshadow's use case. I could add a check to make sure that the source is a directory (and error otherwise).

activeshadow:
@jcrussell @djfritz volume mounts are working for me!

So. Very. Awesome.

Thanks!

PR #894, opened 2017-03-17
gofmt passwordify

floren:
Making the build fail on un-gofmt'd code, in my opinion, makes us
even worse than Python: rather than fail due to significant whitespace,
we fail due to insignificant whitespace.

#892

### DIFF
--- a/src/passwordify/main.go
+++ b/src/passwordify/main.go
@@ -18,7 +18,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


jcrussell:
We should enable the travis integration -- then merges will be blocked until the code is run through gofmt.

floren:
Agreed, travis integration is great.

PR #897, opened 2017-03-17
minimega: add `vm config miniccc`

### LINKED ISSUE #896: Feature Request: vm config serialcc false
By default mini adds a serial interface for cc. I have a vm that I need a serial adapter on and crashes when presented two. 

As a workaround I've been usIng qemu-override but this isn't ideal since the serial port is created using the id # and afaik I can't regex string replace. So I have to know what the vm id # is gonna be before string replacing it, thankfully it just increments. 

A vm config kvmmod false to remove -enable-kvm would help too when I change the qemu binary to arm.

jcrussell:
Enable/disable cc serial port for a VM. Tested with container and KVM.

Fixes #896


djfritz:
The API name is confusing because you can still use miniccc over TCP even
with this disabled.

jcrussell:
Any suggestions? I'm not too fond of any of `cc-serial`, `serial-cc`, `miniccc-serial`, or `serial-miniccc`.

mkunz7:
I think I like miniccc-serial the most. Here's a list I thought of.
```
activecoms (as in com1,com2,com3 in windows)
comcc
ttycc
cctty (play on putty)
minicom (play on the program to connect to a serial port)
serial-console
cereal (I'm hungry)
rs232cc
232cc
uartcc
serialbackdoor
serialbot
serialchannel
ccchannel
```

jcrussell:
```
cereal (I'm hungry)
```
+1, we need more puns.

floren:
I lean toward `vm config serial-cc` or something similar, because as fritz points out this doesn't affect TCP.

I also think there's one too many c's in miniccc but that's for another time.

djfritz:
@jcrussell what's the determination on this?

jcrussell:
How about `vm config backchannel`?

If we like `vm config serial-cc <true,false>`, I can change it to that.

floren:
vm config backchannel would work for me.

jcrussell:
@mkunz7: can you test this and see if it works for you.

mkunz7:
Head on the pr does not have your local branch changes. Using your branch, I'm still getting `-device virtio-serial` in `qemu-override` with `backchannel false`. Can we get rid of this? My vm crashes if I don't axe it with `qemu-override`.

Also do we want to add vm config backchannel to vm config printing?

jcrussell:
@mkunz7: trimmed the extra `-device`, hopefully this works for you.

mkunz7:
```
git clone https://github.com/jcrussell/minimega.git
git checkout issue-896
git checkout 9445f42fc25192f12c6e4f15c465b77addc1254b
./all.bash
```

LGTM

PR #898, opened 2017-03-22
igor: replace sync/ping with nmap

### LINKED ISSUE #881: igor: faster up/down checking
igor sends 1 ping to each node to check if it is up or down. This means 'igor show' can take several seconds to display if most of the cluster is down. We should be able to speed it up.

jcrussell:
Way faster and simpler code. Fixes #881.

### DIFF
--- a/src/igor/show.go
+++ b/src/igor/show.go
@@ -8,7 +8,6 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


jcrussell:
Old igor, one host down:
```
real    0m1.014s
user    0m0.012s
sys     0m0.036s
```

New igor, one host down:
```
real    0m0.123s
user    0m0.012s
sys     0m0.012s
```

PR #902, opened 2017-03-30
bridge: ``fix'' qos code

jcrussell:
Use tbf instead of netem for rate limiting since netem didn't seem to
behave correctly between VMs on different hosts. Unfortunately, this
means that rate and loss/delay are mutually exclusive now. Documented
this and made a few cleanups. Updated tests.

Tested between two containers on different hosts.


jasoncontrib:
LGTM

PR #904, opened 2017-04-10
Igor improvements

floren:


floren:
Do not merge this yet.

floren:
So this has turned into everything **except** batch operations for igor, despite the PR name. Here's what's new:

- Reservations are scheduled in advance, so if the cluster's full you'll get the next available slot
- You can specify a time in the future for your reservation, rather than next available
- You can finally request any N nodes, rather than manually specifying what nodes you want
- Cobbler integration for netbooting (with the old method still available)
- Optional automatic powercycling at reservation start and finish.
- Optional automatic Q-in-Q network segmentation per reservation (Arista switches only)

PR #906, opened 2017-04-17
ron/minimega: huge refactoring for namespaces

jcrussell:
We now start one ron server per namespace:

 * remove all occurrences of namespace from minicli/ron -- connecting
   VMs with the correct server is the responsibility of minimega
 * allow clients with the same UUID to connect (assuming that they are
   in different namespaces)
 * separated responses for different namespaces into subdirectories
 * reads files first from the namespace specific directory and then the
   "global" directory

There was also refactoring to move functionality into the ron package so
that we can simply call functions on the correct ron.Server instance.

Also, got rid of prefixMap in favor of embedding the prefix in commands.


jcrussell:
I'm going to roll this into my larger namespace refactoring.

PR #913, opened 2017-05-04
minimega/minicli: improved env support

### LINKED ISSUE #909: minicli: add `.env`
Add:
```
.env [name]
.env <name> <value>
```

To inspect and update the env variables.

jcrussell:
Added .env to print/update/unset env variables. Use os.ExpandEnv instead
of custom code to expand env variables -- now support $var or ${var}.

Closes #909


djfritz:
We used to have https://github.com/djfritz/gomacro to provide some of this (including some cool regexp expansion).

Issue #717, opened 2016-10-24
minimega: separate `vmConfig` per namespace


jcrussell:
We currently use one global variable to store the `vm config` state. This won't work if we're trying to describe two experiments at the same time. Should create a copy of vmConfig per namespace.

jcrussell:
Fixed in #915.

PR #915, opened 2017-05-05
minimega: namespace refactoring

### LINKED ISSUE #840: doc: namespaces tutorial
Add article for namespaces tutorial.

jcrussell:
This is a huge refactoring of how namespaces work. It has been tested on my machine and most tests pass. A lot more testing needs to be done. The main idea was to store all state in the `Namespace` struct and pass it in to the different `cli` handlers. This means that each component (e.g. `VMs`, `vnc`, `capture`, ...) doesn't need to be namespace-aware.

Major changes (hopefully this is everything, please complain if you disagree with any):

- Always in a namespace (`minimega` by default).
- You can't issue commands across namespaces (e.g. `vm kill all`) -- you can only affect the active namespace.
- There are separate `vm config`s per-namespace (#717). The active `vm config` is not copied when creating a new namespace.
- Each namespace has its own `ron.Server` and registers VMs with the correct server.
- Each `ron.Server` writes responses into its own `miniccc_responses` directory.
- `nsmod` was renamed to `ns` and has a lot more functionality (e.g. `ns hosts`, `ns queue`, `ns flush`, `ns schedules`).
- When launching VMs, they will be queued or launched immediately based on `ns queueing`. Default is not to queue.
- Eliminated `noblock` option for `vm launch` -- equivalent to queueing.
- `clear namespace <name>` now destroys state associated with namespace including killing VMs.
- There is a separate `cc filter` and `cc prefix` per namespace.
- Removed `ccport` flag and added `cc listen` API to make the cc server listen on a port.
- There is a separate `router` config per namespace.
- Running `namespace` produces the list of namespaces instead of displaying info about the current namespace (although the current namespace is denoted in the list).
- `optimize` is partly namespace aware (this probably needs tweaking).
- `mesh send all` *always* means all meshage-connected nodes (reverts change from 2.3).
- Add `tap` and `bridge` completion.
- Add distributed minitests using miniception.

Things that are not namespaced:

- meshage
- history
- dnsmasq
- netflow timeouts
- VM IDs
- logging

TODO:

- [x] Fix #840 
- [x] Test on cluster
- [x] Tweak scheduler -- needs VMs across namespaces
- [x] Fix minitests
- [x] Fix distributed minitests
- [x] Fix developer namespaces article


jcrussell:
I've been using this a little but not extensively enough to call it tested. I have some ideas for improving minitest to help test this that I'll start to work on. I'm hoping that once we're sufficiently happy with the minitests that we can merge and triage bugs as they appear before the 2.4 release.

jcrussell:
"Tweak scheduler -- needs VMs across namespaces" -- should we actually do this? Coschedule could just be within the namespace.

Perhaps we could add a "dibs" API to make a host exclusive to a namespace.

@djfritz

jcrussell:
Found out why the `router` test is failing. I built `minirouterfs` on debian testing and it looks like there's a problem with `dhclient-script`:

```       
DHCPDISCOVER on veth0 to 255.255.255.255 port 67 interval 4                     
DHCPREQUEST of 1.2.3.135 on veth0 to 255.255.255.255 port 67                    
DHCPOFFER of 1.2.3.135 from 1.2.3.4                                             
DHCPACK of 1.2.3.135 from 1.2.3.4                                               
/sbin/dhclient-script: line 418: can't open /etc/fstab: no such file            
DHCPDECLINE on veth0 to 255.255.255.255 port 67                                
DHCPDISCOVER on veth0 to 255.255.255.255 port 67 interval 5                     
DHCPREQUEST of 1.2.3.136 on veth0 to 255.255.255.255 port 67                    
DHCPOFFER of 1.2.3.136 from 1.2.3.4                                             
DHCPACK of 1.2.3.136 from 1.2.3.4                                               
/sbin/dhclient-script: line 418: can't open /etc/fstab: no such file            
DHCPDECLINE on veth0 to 255.255.255.255 port 67                                
...
```

This repeats about 250 times to fill the /24. Perhaps we should vendor the `dhclient-script` in the `minirouterfs` build script?

jcrussell:
Related issue: https://bugs.launchpad.net/ubuntu/+source/libguestfs/+bug/1632405

jcrussell:
Found a possible bug with containers today, need to triage:

```
vm kill <container>
vm start <container>
```

and `/cc` doesn't get recreated.

jcrussell:
Ran all minitests 100x without a crash or a failure.

jcrussell:
@floren @djfritz: PTAL.

jcrussell:
Confirmed that minitests still pass and tested on the cluster.

PR #922, opened 2017-05-26
miniweb: clean up code, set record false, add `vm top` view

jcrussell:
Use sort.Slice to clean up some code. Made vmInfo cleaner and use in the
vms handler. Set record false so we don't clutter the history.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ dist: trusty
 sudo: false
 language: go
 go:
-  - 1.7.3
-  - 1.8
+  - 1.8.3
 addons:
   apt:
     packages:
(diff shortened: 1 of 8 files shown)


jcrussell:
Failed tests because `sort.Slice` isn't in Go 1.7.

djfritz:
Does travis have 1.8 yet? I have a hard time merging this even knowing it works because every PR moving forward will fail the tests...

djfritz:
@jasoncontrib

jasoncontrib:
It did pass in travis with Go 1.8: https://travis-ci.org/sandia-minimega/minimega/jobs/236451290

We probably want to remove 1.7.x from our travis config if we merge this, though, since that'll fail the whole travis build.

jcrussell:
@djfritz -- changed travis config to remove 1.7 from tests.

jcrussell:
@djfritz: PTAL, this should be ready to merge.

jcrussell:
@jasoncontrib: PTAL at the `vm top` view.

PR #924, opened 2017-05-31
noVNC Mobile

mkunz7:
updated old pr to latest


mkunz7:
old pr https://github.com/sandia-minimega/minimega/pull/859

misc/web/novnc/include/ui.js
```diff
@@ -1260,7 +1260,7 @@ var UI;
 
         // Display the desktop name in the document title
         updateDocumentTitle: function(rfb, name) {
-            document.title = name + " - noVNC";
+            //document.title = name + " - noVNC";
```
floren: what's up with this?

floren:
Tested, seems fine. @djfritz @jcrussell any complaints?

misc/web/novnc/include/ui.js
mkunz7: vnc_auto.html sets the window title to the vm name, iirc this gets overwritten when calling UI.connect() so I commented out the update

PR #932, opened 2017-06-29
DHCP Servers Integration

xchenan:
Integrated DHCPv4 and DHCPv6 servers in minimega. 

Dnsmasq that we've been using in minimega has flaws: 
* Dnsmasq commands in different distro can be different.
* Dnsmasq command line in minimega does not support IPv6.

DHCPv4 server is a credit to Richard Warburton: https://github.com/krolaw/dhcp4
DHCPv6 server is a credit to Matt Layher: https://github.com/mdlayher/dhcp6


floren:
The entirety of the minimega src/ tree currently only comprises about 170,000 lines of code. Asking to add another 144,000 LOC sticks in my craw.

src/minimega/dhcp6server_cli.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/mdlayher/dhcp6"
+	"golang.org/x/net/ipv6"
+)
+
+var dhcp6CLIHandlers = []minicli.Handler{
+	{ // dhcp6Server
+		HelpShort: "start a dhcpv6 server which serves a specified ip",
+		HelpLong: `
+Start a dhcp/dns server that serves a specified IP. For example,
```
floren: All this code to serve 1 IP? Just use Router Advertisement imo.

rminnich:
well, look at it this way. You're currently dependent on dnsmasq and we've found that isn't working well, as dnsmasq is not portable across all the varieties of linux. Further, the dnsmasq dependency complicates building minimega IoT devices. And, finally, this is vendored code, not minimega code. I realize it's a giant wad of code, but ... it does remove the dnsmasq dependency. If you don't want to do this we'll work out something else, however. We can do a simple dhcpserver in go for u-root for example.

jcrussell:
This is pretty neat! Instead of using dhcp6, perhaps it would be better for minimega to have a command to start sending router advertisements?

When you say "minimega IoT devices" are you suggesting that you want to use minimega to run VMs (or containers) on IoT devices?

src/minimega/dhcp6server_cli.go
```diff
@@ -0,0 +1,85 @@
...
+	// Only accept a single IPv6 address
+	ip := net.ParseIP(c.StringArgs["ipv6"]).To16()
+	if ip == nil || ip.To4() != nil {
+		return fmt.Errorf("IP is not an IPv6 address")
+	}
+
+	// Set a default interface if not specified
+	iface, ok := c.StringArgs["interface"]
+	if !ok {
+		iface = "eth0"
```
floren: This is not sensible in the context of minimega. A DHCP server should be attached to a host tap, defined using the "tap" commands.

djfritz:
We definitely need to replace dnsmasq bindings, but I think we need to rethink things before heading down another code path to support.

PR #934, opened 2017-07-10
multiple: replace goreadline with liner

jcrussell:
Replaced the c readline bindings with pure go liner. Updated minimega,
rond, and miniclient to use. Changed minipager to use `less` if we get
too many lines (currently twice the terminal height).

Tab completion is more like bash with this PR than #933.


jcrussell:
TODO: we don't properly reset the terminal on `quit`.

jcrussell:
@djfritz @floren: PTAL, I'll close #933 if we like this functionality better.

PR #940, opened 2017-07-19
minimega: fix completion

### LINKED ISSUE #938: minimega: fix filepath completion
The filepath completion isn't very good after #934.

jcrussell:
Fixes issues with completion including env variables and file path
completion. Closes #938.

### DIFF
--- a/src/miniclient/client.go
+++ b/src/miniclient/client.go
@@ -7,6 +7,7 @@ package miniclient
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"minicli"
@@ -40,6 +41,9 @@ type Response struct {
 
 type Conn struct {
 	url string
+
+	// first error encountered
+	err error
 
 	conn net.Conn
 
@@ -81,7 +85,8 @@ func (mm *Conn) Pipe(pipe string) (io.Reader, io.WriteCloser) {
 		PlumbPipe: pipe,
 	})
 	if err != nil {
-		log.Fatal("local pipe gob encode: %v", err)
+		mm.err = fmt.Errorf("local pipe gob encode: %v", err)
+		return nil, nil
 	}
 
 	rr, rw, err := os.Pipe()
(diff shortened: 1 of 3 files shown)


mkunz7:
doesn't work for me

jcrussell:
Better?

jcrussell:
That's strange... for me:

```vm conf<tab>``` ->```vm config ```
```vm config disk /hom<tab>``` -> ```vm config disk /home/```

@floren: Do these work for you?

**Comment by djfritz** [comment]
Appears to work for me - completes "config", completes directories, and completes with directory contents on double-tab.

**Comment by jcrussell** [comment]
@mkunz7 -- have you tried this again? Can we merge?

**Comment by mkunz7** [comment]
Still broken.

```
Ubuntu 16.04.1 LTS
go version go1.8.3 linux/amd64
```

```
git clone http://github.com/sandia-minimega/minimega.git
cd minimega
git pull origin pull/940/head
./all.bash
/home/ubuntu/minimega/bin/minimega -nostdin &
/home/ubuntu/minimega/bin/minimega -attach
vm conf<tab>
config
```
/home/ubuntu/minimega/bin/minimega it's just when you use nostdin and attach

**Comment by mkunz7** [comment]
It's also worth noting an additional bug, when you run quit quit to exit an attached mini console it ruins the ssh session and you will be no longer able to see what you type. Even exiting bash doesn't fix it you have to exit the ssh session.

**Comment by jcrussell** [comment]
That helps, didn't realize it was from the `-attach` interface. I'll take another look.

**Comment by jcrussell** [comment]
You can type `reset` to restore your terminal to a working state. I can look into ensuring that we reset the terminal state on `quit`.

**Comment by jcrussell** [comment]
@mkunz7 -- can you confirm that these are working for you?

**Comment by floren** [comment]
Works right for me, with and without doing -attach.

File: src/miniclient/client.go [hunk 1]
```diff
@@ -81,7 +85,8 @@ func (mm *Conn) Pipe(pipe string) (io.Reader, io.WriteCloser) {
 		PlumbPipe: pipe,
 	})
 	if err != nil {
-		log.Fatal("local pipe gob encode: %v", err)
+		mm.err = fmt.Errorf("local pipe gob encode: %v", err)
```
floren: I didn't see a Lock() prior to writing this; is Pipe() only ever called after something else has locked mm?

jcrussell: Pipe is a bit weird -- it's only invoked when calling `minimega -pipe`. I left it unchanged but we may want to tweak it in the same way later.

PR #948, opened 2017-08-10
vendor: move to govendor

jcrussell:
Use govendor to manage vendorized code. Used versions from commit logs
when available. govendor trims test files so many files will be removed
from the repo.


jcrussell:
Hm, govendor json includes:

```
	"rootPath": "/data/repos/sandia/minimega/src"
```

I wonder if we can change that to `$GOPATH/src`.

jcrussell:
We should probably use [dep](https://github.com/golang/dep) instead -- it might be mainlined in 1.10.

PR #950, opened 2017-08-22
miniweb: add authentication

jcrussell:
Read a json file with bcrypt hashed passwords and use the path
parameters to restrict access. By default, everything is allowed. Added
bootstrap flag to generate password file. For example:

```
$ bin/miniweb -bootstrap -passwords minimega.passwd
Configure /
Username: jon
Password:
Confirm Password:

Add additional users (Ctrl-D when finished):
Path: /connect/fritz
Username: fritz
Password:
Confirm Password:

Add additional users (Ctrl-D when finished):
Path: /connect/john
Username: john
Password:
Confirm Password:
```

This allows you to have a top-level user and refine access to more
specific paths.

Need to fix websockets so that they enforce auth and restructure paths
so that you don't need to grant access to `/connect` and `/tunnel` to
connect to a VM. Probably should move `/tunnel` to `/connect/<vm>/ws`.

Should also add TLS support.


jcrussell:
Think this is ready for review.

jcrussell:
Think this is ready for review.

jcrussell:
Should "/" mean that you can access everything under "/" even if there's a more specific username/password for a path?

For example:
```
jon  /
john /connect/john
```

Should `jon` be able to access `/connect/john`?

floren:
Jon, I would tend to agree that / should give you access to everything.

If we could have namespaces in the path, that would make it a lot easier to restrict access to a subset of VMs.

PR #951, opened 2017-08-22
minimega: add pid to vm info

cg-:
Quick addition of the qemu pid to the vm info output.


floren:
Could we generalize this a bit and print out the host-side PID of a container's init process too? Could be useful for e.g. strace.

cg-:
Sounds good. Added the PID of the container's init process. Right now it's listed twice in the help message since they're slightly different things-- should that be combined?

Issue #709, opened 2016-10-20
minimega: console improvements


jcrussell:
Allow multiple users to connect to the same console. minimega should store a scrollback for context when you connect. Support "screenshot".

jcrussell:
All but screenshot implemented in #954. Not clear that screenshot would be useful.

PR #954, opened 2017-08-22
multiple container terminals & scrollback

floren:
Intended to satisfy #709 

So far multiple terminals open works, and when you open a new terminal it plays back some scrollback so you don't have a blank screen. "screenshot" is going to be very difficult and needs to be punted.


floren:
Can we get this in for minimega 2.4?

Issue #974, opened 2017-09-07
minidoc: clean ups


jcrussell:
* Sidebar should be a separate template so that we don't have to maintain links in multiple places
 * Should have link to presentations
 * Move tutorials to a directory -- it's weird that you click on documentation and see tutorials as a subdir
 * Remove repeated info (Documentation, --articles--, Articles)
 * Make API documents a top-level link
 * Add to `More tools`: `rfbplay`, `vncdrone`, `minitest`, `nfcat`, `miniweb`
 * Fix links on `contributing.article`
 * Add header/sidebars to articles

PR #975, opened 2017-09-08
minidoc/doc: improvements

jcrussell:
Address most of #974. Did not mess with `article.tmpl` to try to get minimega header and links on the article pages.


floren:
This looks good, is it a WIP or ready to merge? I've tested it locally and it seems fine.

jcrussell:
I think this is good for now -- messing with `article.tmpl` so that it displays the header and sidebar seems like too much work for now.

PR #976, opened 2017-09-08
minimega: add mkdirMount

### LINKED ISSUE #957: minimega: container code, make guest vm /dev /proc /sys etc if they don't exist
see summary

jcrussell:
Add helper to mkdir and then mount for containers. Fixes #957.


floren:
the diff is really a nightmare, but this looks pretty ok to me.

PR #983, opened 2017-09-20
some final igor updates

floren:
Update igor docs

Allow reservations to specify a cobbler profile instead of kernel+initrd.


floren:
Seems to be working well enough to merge, after a few days testing on a real cluster.

Issue #624, opened 2016-08-20
minimega: hanging namespaces


djfritz:
It's possible to hang minimega over meshage with namespaces:

node1: enter a namespace
node2: mesh send all something like vm info

node1, depending on the command, may in turn attempt to mesh send commands, causing a deadlock.

This doesn't need a fix right now, but we should get to it eventually.

@jcrussell

jcrussell:
I thought we decided two head nodes was bad.

We could fix this by making it so that we only ever invoke the local behavior when we receive a command via meshage.

jcrussell:
Whoops...

jcrussell:
We can probably solve this with the `minicli.Command.Source` field. We already use this field to determine whether to do the "fan out" or "local" behavior in `wrapBroadcastCLI` and `wrapVMTargetCLI`. We would need to change these so that they perform the local behavior whenever the Source is a non-zero string. We would also need to change `mesh send` to set the source of the subcommand to a non-zero string.

PR #985, opened 2017-09-20
minimega: destroy namespace on remote nodes

### LINKED ISSUE #984: minimega: `clear namespace` for remote hosts
`clear namespace <namespace>` should destroy the namespace on the local node and the remote nodes.

Right now, it only destroys it on the local node which leaves VMs running on the remote node unless you use:

```
mesh send all clear namespace <namespace>
clear namespace <namespace>
```

jcrussell:
Fixes #984 by broadcasting out the `clear namespace <namespace>` command
to all remote nodes. Very prone to triggering #624.


doc/content/articles/developer/namespaces.article
```diff
@@ -59,5 +59,5 @@ Without some mechanism to resolve this, we would fan out again and cause a
 deadlock. To prevent this, we tag the outgoing `minicli.Command` using the
 `Source` field. Specifically, we set the `Source` field to the active
 namespace. The `wrapBroadcastCLI` and `wrapVMTargetCLI` handlers check the
-`Source` field against the active namespace and, if it matches, perform the
-local behavior. Otherwise, they will fan out.
+`Source` field and, if it non-zero, perform the local behavior. Otherwise, they
```
floren: you accidentally a word before 'non-zero'

jcrussell:
#624 is not fully fixed -- if you `mesh send <host> vm launch`, we'll deadlock.

jcrussell:
You can also break things with `ns run ns run ...`.

jcrussell:
This is similar to `mesh send all mesh send ...` -- we disallow the `mesh send` subcommand in `mesh send`. We would have to change `mesh send` to check for `ns run` as well. Perhaps it would be better to move the check to the subcommand -- some commands cannot be subcommands (e.g. `mesh send` and `ns run`).

glattercj:
Seems to be fixed on our end.

Issue #876, opened 2017-02-21
miniweb: forced namespace


jcrussell:
Add a `-namespace` flag to limit access to a particular namespace.

Issue #873, opened 2017-02-21
miniweb: upload/download files


jcrussell:
Add functionality to upload/download files to/from the `iomeshage` directory.

mkunz7:
I like how this guy did it, writing an api interfacing angular-filemanager

https://github.com/kernel164/gofe

csymonds:
This has been incorporated into the new miniweb work #1047

Issue #872, opened 2017-02-21
miniweb: VM controls


jcrussell:
Add buttons to start/stop/kill/... VMs.

Issue #870, opened 2017-02-20
miniweb: add namespace as optional path parameter


derhart:
We have experiments that we deploy in separate namespaces that have the same virtual machine name. For instance, there is a server1 in namespace foo and a server1 in namespace bar. The VNC web path resolves to http://<ip>:9001/connect/server1 for both, and this is making it impossible to connect to one of the boxes. Can you update the path to be namespace aware?

derhart:
This looks to be intermittent, so I'm wondering if there's a race condition coming into play. I haven't found a clear pattern to what causes it though.

jcrussell:
This is one of the features we want in miniweb. The current web will connect to whatever VM is in the active namespace of the head node.

jcrussell:
@djfritz -- should we merge #852 and set a timeline for adding this?

djfritz:
We can. Perhaps we should change the behavior to use the current namespace if none is specified (current behavior), and allow for a larger URL for namespace specific conns:

    # connect to server1 in the current namespace
    http://.../connect/server1

    # connect to server1 in a specific namespace
    http://.../connect/foo/server1

Thoughts?

@jcrussell

djfritz:
Later on we can enforce namespace-specific constraints in miniweb with a -namespace flag or something, and that could disable the above feature.

jcrussell:
I think adding the namespace as a path parameter makes sense and it wouldn't be simple to add.

PR #987, opened 2017-09-24
miniweb: braver and newer web

jcrussell:
minimega:
 - display more info about namespaces in `namespace` API

miniweb:
 - force namespace with -namespace flag:
```
    /foo/vms gives you VMs from foo. /vms gives you VMs from the current
    namespace. /foo/bar/vms gives you an error.
```
 - add namespaces page: displays info about different namespaces
 - add files pages: directory walk of files for hosts in the namespace
 - add vlans page: displays VLAN aliases
 - add VM controls: start/stop/kill VMs

Fixes a few issues: #870, #872, #876.
Partially implements #873.


jcrussell:
Small bug: `namespace` counts VMs on the current host and not across the namespace.

PR #991, opened 2017-09-26
doc: add miniweb article

### LINKED ISSUE #990: miniweb: article
miniweb has grown complex enough that we probably need an article about it.

jcrussell:
Fixes #990.

### DIFF
--- a/doc/content/articles/miniweb.article
+++ b/doc/content/articles/miniweb.article
@@ -1,0 +1,177 @@
+miniweb
+
+The minimega authors
+26 Sept 2017
+
+* Introduction
+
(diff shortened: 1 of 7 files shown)


doc/content/articles/miniweb.article
```diff
@@ -0,0 +1,177 @@
+miniweb
+
+The minimega authors
+26 Sept 2017
+
+* Introduction
+
+`miniweb` is a webserver for minimega based on the `web` API that existed until
```
floren: I think it would be good to mention why we split miniweb out of the minimega binary... this sentence confuses me a little.

Has all the documentation been updated to reflect that we now use miniweb instead of web?

PR #996, opened 2017-10-10
igor: fix bugs, add power control

floren:
Fixes some bugs with how reservations worked in a Cobbler environment,
fixes a bug in allocating VLANs.
Adds the ability to power nodes on and off.
Adds additional logging to a file.
Adds limits to how many nodes a reservation can contain and how long
a reservation can run.


jcrussell:
Is this still a WIP?

floren:
I'll do a last look tomorrow but I think it's good to go.

floren:
PTAL

PR #1001, opened 2017-10-20
minimega: improve vm net

### LINKED ISSUE #994: minimega: improve `vm net`
Could easily allow it to support "VM target" like `vm start`. Would be nice to do:

```
vm net disconnect client[0-9] 0
```

or

```
vm net connect client[0-9] 0 mega_bridge LAN
```

We should probably change the order of the arguments too so that bridge is last and optional (defaults to `mega_bridge` like everywhere else).

jcrussell:
`vm net` now supports the same vm target syntax as `vm start`. Made
bridge an optional parameter that defaults to `mega_bridge`. Fixes bug
where `vm net` used wrapSimpleCLI instead of wrapVMTargetCLI -- before
it wouldn't work across a namespace.

Fixes #994.

We should merge this before 2.4 since it includes a bug fix (and a
useful feature). Most of the additions are tests.

### DIFF
--- a/src/minimega/vm.go
+++ b/src/minimega/vm.go
@@ -519,7 +519,17 @@ func (vm *BaseVM) NetworkConnect(pos int, bridge string, vlan int) error {
 
 	net := &vm.Networks[pos]
 
-	log.Debug("moving network connection: %v %v %v -> %v %v", vm.ID, pos, net.VLAN, bridge, vlan)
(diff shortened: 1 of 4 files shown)


jcrussell:
TODO: running `connect` should reuse the same bridge if not specified and not use `mega_bridge`.

PR #1007, opened 2017-10-27
Igor: Fixes and Updates

csymonds:
This is a check-in for the following igor issues:
1. using both -w and -n should result in some kind of error and not be ignored
2. Reservations were being made slightly <=1min into the future
3. power.go - wrong log call used
4. Adding in time units to -t allowing # days specified

### DIFF
--- a/src/igor/sub.go
+++ b/src/igor/sub.go
@@ -141,7 +141,7 @@ func runSub(cmd *Command, args []string) {
 		}
 	}
 
-	when := time.Now()
+	when := time.Now().Add(-time.Minute * MINUTES_PER_SLICE) //keep from putting the reservation 1 minute into future
 	if subA != "" {
 		loc, _ := time.LoadLocation("Local")
 		t, _ := time.Parse(format, subA)
--- a/src/igor/sub.go
+++ b/src/igor/sub.go
@@ -103,7 +103,7 @@ func runSub(cmd *Command, args []string) {
 			log.Fatal("Unable to parse -t argument: %v\n", err)
 		}
 	}
-	fmt.Printf("duration: %v, nano: %v\n", duration, nanoseconds/time.Minute)
+	log.Debug("duration: %v, nano: %v\n", duration, nanoseconds/time.Minute)
 	duration = duration + int(nanoseconds/time.Minute)
 	if duration < MINUTES_PER_SLICE { //1 slice minimum reservation time
 		duration = MINUTES_PER_SLICE
--- a/src/igor/sub.go
+++ b/src/igor/sub.go
@@ -171,10 +171,13 @@ func runSub(cmd *Command, args []string) {
 		return
 	}
 
-	if subN > 0 {
-		reservation, newSched, err = findReservationAfter(subT, subN, when.Unix())
-	} else if subW != "" {
+	if subW != "" {
+		if subN > 0 {
+			log.Fatalln("Both -n and -w options used. Operation canceled.")
+		}
 		reservation, newSched, err = findReservationGeneric(subT, 0, nodes, true, when.Unix())
+	} else if subN > 0 {
+		reservation, newSched, err = findReservationAfter(subT, subN, when.Unix())
 	}
 	if err != nil {
 		log.Fatalln(err)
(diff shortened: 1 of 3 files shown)


src/igor/sub.go
```diff
@@ -69,7 +71,7 @@ func init() {
 	cmdSub.Flag.StringVar(&subI, "i", "", "")
 	cmdSub.Flag.IntVar(&subN, "n", 0, "")
 	cmdSub.Flag.StringVar(&subC, "c", "", "")
-	cmdSub.Flag.IntVar(&subT, "t", 60, "")
+	cmdSub.Flag.StringVar(&subT, "t", "60m", "")
```
jcrussell: Too bad we want to support days otherwise we could just use `flag.Duration`. Here's why `time.ParseDuration` does not support days: https://github.com/golang/go/issues/11473.

src/igor/sub.go
```diff
@@ -41,7 +42,7 @@ OPTIONAL FLAGS:
 
 The -c flag sets any kernel command line arguments. (eg "console=tty0").
 
-The -t flag is used to specify the reservation time in integer minutes. (default = 60)
+The -t flag is used to specify the reservation time. Time denominations should be specified in days(d), hours(h), and minutes(m), in that order. Example: To make a reservation for 7 days: -t 7d. To make a reservation for 4 days, 6 hours, 30 minutes: -t 4d6h30m (default = 60m)
```
jcrussell: Add a note about a day being defined as 24*60 minutes?

src/igor/sub.go
```diff
@@ -82,6 +84,33 @@ func runSub(cmd *Command, args []string) {
...
+				if subT != "" {           // capture any additional time
+					nanoseconds, err = time.ParseDuration(subT)
+				}
+			}
+		}
+		if err != nil {
+			log.Fatal("Unable to parse -t argument: %v\n", err)
+		}
+	}
+	fmt.Printf("duration: %v, nano: %v\n", duration, nanoseconds/time.Minute)
```
jcrussell: Should probably `log.Debug` this.

src/igor/sub.go
```diff
@@ -82,6 +84,33 @@ func runSub(cmd *Command, args []string) {
...
 	format := "2006-Jan-2-15:04"
 
+	// parse duration requested
+	var days int = 0
+	duration = 0
+	nanoseconds, err := time.ParseDuration(subT)
+	if err != nil {
+		// Check for a number of days in the argument
+		var dInd int
+		if dInd = strings.Index(subT, "d"); dInd >= 0 {
```
jcrussell: If a variable is only used in the `if`, you can declare it inline:

`if dInd := strings.Index(subT, "d"); dInd >= 0 {`

src/igor/sub.go
```diff
@@ -54,11 +55,12 @@ var subK string       // -k flag
 var subI string       // -i
 var subN int          // -n
 var subC string       // -c
-var subT int          // -t
+var subT string       // -t
 var subS bool         // -s
 var subA string       // -a
 var subW string       // -w
 var subProfile string // -profile
+var duration int      // parsed subT Duration in minutes
```
jcrussell: Does this need to be a global? Looks like it's passed around as an argument.

csymonds: Yeah, wow. Not sure what my thought process there was. Fixed.

src/igor/sub.go
csymonds: Yeah I read that. At the end of the day it smacks to me like an arbitrary cut-off that they simply don't want to budge from, given the leap-second argument still applies at the hour increment. I'm not too worried as we repeat back the reservation end time which the user can confirm to be correct, however you are right that it should be expressly noted in the help text.

src/igor/sub.go
csymonds: Old habits :-)

src/igor/sub.go
csymonds: log.Fatal should catch this, and Fatal seems like the appropriate action here. Let's discuss this one when you are back.

jcrussell: I meant for the `fmt.Printf`.

Issue #1000, opened 2017-10-19
igor: add `igor version`


jcrussell:
Similar to `version` in minimega -- print out `version.Revision`.

PR #1014, opened 2017-10-27
igor: add version option

csymonds:
Adding the option to print version information for igor


csymonds:
This pr will complete issue #1000

PR #1029, opened 2017-11-21
minimega: rewrite vnc playback

jcrussell:
Fixes a bunch of potential issues with the vnc playback code including leaked file descriptors and race conditions. Updates #1021.

@mkunz7: could you try this with some of your VNC-heavy experiments?


mkunz7:
My vnc heavy experiments use an old version of mini before the rewrites. They did so in a clumsy matter where timing between clicks would matter a lot more. For how VNC is used now and for the foreseeable future this works and fixes the negative time issue.

PR #1032, opened 2017-11-27
igor: stricter usage of `cobbler profile list`

jcrussell:
Parse `cobbler profile list` into a map and check whether the requested
profile is in the map. This prevents a substring of a valid profile name
from being used.


csymonds:
lgtm - need to merge manually

PR #1037, opened 2017-12-08
pyapigen: tool to generate python bindings

jcrussell:
_WIP: Don't merge, yet._

Based on @devinc's genapi.py but implemented in Go. Bindings may need some
tweaking but this simple example works:

```python
import minimega

mm = minimega.connect(debug=False)

print 'launching vms'
mm.vm_launch_kvm(10)

print 'vm info:'
minimega.print_rows(mm.vm_info())

print 'starting vms'
mm.vm_start('all')

print 'vm info:'
minimega.print_rows(mm.vm_info())

print 'killing vms'
mm.vm_kill('all')

print 'flushing vms'
mm.vm_flush()
```


jcrussell:
TODO:
```python
with minimega.namespace("name") as mm:
    mm.vm_info()
```

PR #1061, opened 2018-01-23
igor: Adding Bounds Check

csymonds:
Adding in a bounds check to prevent a panic when user requests a node out of range.

### DIFF
--- a/src/igor/power.go
+++ b/src/igor/power.go
@@ -129,6 +129,11 @@ func runPower(cmd *Command, args []string) {
 		if len(nodes) == 0 {
 			log.Fatal("Couldn't parse node specification %v\n", subW)
 		}
+		// make sure the range is valid
+		if inRange, err := checkValidNodeRange(nodes); !inRange {
+			log.Fatal("Invalid node range: %v", err)
+		}
+
 		// This will be the list of nodes to actually power on/off (in a user-owned reservation)
 		var validatedNodes []string
 		for _, n := range nodes {
@@ -139,7 +144,7 @@ func runPower(cmd *Command, args []string) {
 				log.Fatal("choked on a node named %v", n)
 			}
 
-			resID := currentSched.Nodes[index]
+			resID := currentSched.Nodes[index-1]
 			for _, r := range Reservations {
 				if r.ID == resID && r.Owner == user.Username {
 					// Success! This node is in a reservation owned by the user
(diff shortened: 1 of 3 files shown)


src/igor/schedule.go
```diff
@@ -44,6 +44,9 @@ func areNodesFree(clusternodes []uint64, requestedindexes []int) bool {
 
 // Returns true if nodes[index] through nodes[index+count-1] are free (set to 0)
 func isFree(nodes []uint64, index, count int) bool {
```
jcrussell: While this solves the panic, does it make sense to put the check this deep in the code? Shouldn't `runSub` do the check?

src/igor/power.go
```diff
@@ -139,7 +144,7 @@ func runPower(cmd *Command, args []string) {
 				log.Fatal("choked on a node named %v", n)
 			}
 
-			resID := currentSched.Nodes[index]
+			resID := currentSched.Nodes[index-1]
```
jcrussell: This is the off-by-one error? `currentSched` is 0 based?

src/igor/schedule.go
```diff
@@ -11,6 +11,18 @@ import (
 	"time"
 )
 
+func checkValidNodeRange(nodes []string) (bool, error) {
```
jcrussell: This should return bool and if there's an error, return false.

src/igor/power.go
csymonds: Yep. There is likely a better way to do this, but that might take a more comprehensive approach. For now I'm just fixing the error.

src/igor/schedule.go
```diff
@@ -11,6 +11,14 @@ import (
 	"time"
 )
 
+func checkValidNodeRange(nodes []string) bool {
+	indexes, err := getNodeIndexes(nodes)
+	if err != nil {
+		log.Fatal("Unable to get node indexes: %v", err)
```
jcrussell: We should avoid calling `log.Fatal` and instead return false here.

Most of the time, we will want to call `log.Fatal` if the node check fails but we don't want to not use the function because it could cause the program to exit. It will also make it easier to write a couple of test cases ;) ;).

csymonds: Because we are no longer returning err and just returning a generic boolean, I think having something that will more easily distinguish why we failed here will be helpful.

PR #1062, opened 2018-01-24
igor: Adding reservation extension feature

mpopple:
- Can extend existing reservations by a specified amount of time (with the same format as igor sub)


src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+
+	cmdExtend.Flag.StringVar(&subR, "r", "", "")
+	cmdExtend.Flag.StringVar(&subT, "t", "60m", "")
+}
+
+func runExtend(cmd *Command, args []string) {
+	// duration is in minutes
+	duration := 0
+
+	v, err := strconv.Atoi(subT)
```
jcrussell: Pull this out into a function and call it from `runSub` and `runExtend` rather than copy it.

```
// parseDuration parses a duration, supporting a 'd' suffix in addition to those supported by time.ParseDuration.
// Returns the duration in minutes.
func parseDuration(s string) (int, error) {
...
}
```

src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+		if duration > igorConfig.TimeLimit {
+			log.Fatal("Only root can make a reservation longer than %v minutes", igorConfig.TimeLimit)
+		}
+	}
+
+	// Make sure there's already a reservation with this name
+	for _, r := range Reservations {
+		if r.ResName == subR {
+			if r.Owner != user.Username {
+				log.Fatal("Insufficient permissions for accessing reservation %v", subR)
```
jcrussell: This should probably say that you are not the owner of this reservation instead.

src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			// Check to see if nodes are free to extend
+			hosts := r.Hosts
+			for _, s := range Reservations {
```
jcrussell: Maybe call it `r2` instead of `s` to make it easier to follow.

src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+			fmt.Printf("Nodes: %v\n", unsplit)
+
+			emitReservationLog("EXTENDED", r)
+
+			break
+		}
+	}
+
+	//TODO: Call new Sched
+	//Schedule = newSched
```
jcrussell: Probably needed.

src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+
+	timefmt := "Jan 2 15:04"
+	for _, r := range Reservations {
+		fmt.Printf("%v, %v\n", r.ResName, time.Unix(r.EndTime, 0).Format(timefmt))
```
jcrussell: Don't think we need to print all the reservations here.

src/igor/extend.go
```diff
@@ -0,0 +1,143 @@
...
+
+	user, err := getUser()
+	if err != nil {
+		log.Fatalln("cannot determine current user", err)
+	}
+
+	// Make sure the reservation doesn't exceed any limits
+	if user.Username != "root" && igorConfig.TimeLimit > 0 {
+		if duration > igorConfig.TimeLimit {
+			log.Fatal("Only root can make a reservation longer than %v minutes", igorConfig.TimeLimit)
```
jcrussell: Is this the right calculation? We stop them from extending it by more than `igorConfig.TimeLimit` but they could extend by 10 minutes 100x.

src/igor/extend.go
mpopple: Where should the resultant function go? In main.go?

src/igor/extend.go
mpopple: Thanks for catching that error. Should be "float64(duration) + r.Duration > float64(igorConfig.TimeLimit)" with r being the existing reservation. Also will change "make a reservation" to "extend a reservation" in the error message.

src/igor/extend.go
jcrussell: Either `time.go` or `misc.go`.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+	user, err := getUser()
+	if err != nil {
+		log.Fatalln("cannot determine current user", err)
+	}
+
+	// Make sure there's already a reservation with this name
+	exists := false
+
+	for _, r := range Reservations {
+		if r.ResName == subR { // The reservation name is unique
```
jcrussell: Change this to `!=` and continue so that the code below is less nested.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+		log.Fatalln("Missing required argument")
+	}
+
+	user, err := getUser()
+	if err != nil {
+		log.Fatalln("cannot determine current user", err)
+	}
+
+	// Make sure there's already a reservation with this name
+	exists := false
```
jcrussell: Don't need this variable... see below.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+
+			timefmt := "Jan 2 15:04"
+			rnge, _ := ranges.NewRange(igorConfig.Prefix, igorConfig.Start, igorConfig.End)
+			fmt.Printf("Reservation %v extended to %v\n", r.ResName, time.Unix(r.EndTime, 0).Format(timefmt))
+			unsplit, _ := rnge.UnsplitRange(r.Hosts)
+			fmt.Printf("Nodes: %v\n", unsplit)
+
+			emitReservationLog("EXTENDED", r)
+
+			exists = true
```
jcrussell: Call `putReservations` and `putSchedule` here and then return.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+
+			break
+		}
+	}
+
+	// If the reservation does not exist then we error out
+	if !exists {
```
jcrussell: Should always be fatal now.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+		log.Fatal("Reservation %v does not exist", subR)
+	}
+
+	// Update the reservation file
+	putReservations()
```
jcrussell: Moved.

src/igor/sub.go
```diff
@@ -13,8 +13,8 @@ import (
 	"os"
 	"path/filepath"
 	"ranges"
-	"strconv"
-	"strings"
+//	"strconv"
```
jcrussell: Delete.

src/igor/extend.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		log.Fatal("unable to parse -t: %v", err)
+	} else if duration < MINUTES_PER_SLICE { //1 slice minimum reservation time
+		log.Fatal("Please specify an extension of at least %v minute(s) in length.", MINUTES_PER_SLICE)
+		//duration = MINUTES_PER_SLICE
+	}
+	log.Debug("duration: %v minutes", duration)
+
+	// Validate arguments
+	if subR == "" || subT == "" {
```
jcrussell: Only need to check `subR` here since `subT` has already been parsed.

jcrussell:
LGTM

jcrussell:
Make sure you run `gofmt` on your code.

src/igor/extend.go
```diff
@@ -0,0 +1,121 @@
...
+	// break init cycle
+	cmdExtend.Run = runExtend
+
+	cmdExtend.Flag.StringVar(&subR, "r", "", "")
+	cmdExtend.Flag.StringVar(&subT, "t", "60m", "")
+}
+
+func runExtend(cmd *Command, args []string) {
+	// duration is in minutes
+	duration := 0
```
jcrussell: Delete.

src/igor/extend.go
```diff
@@ -0,0 +1,121 @@
...
+
+		// Make sure the reservation doesn't exceed any limits
+		if user.Username != "root" && igorConfig.TimeLimit > 0 {
+			if float64(duration)+r.Duration > float64(igorConfig.TimeLimit) {
+				log.Fatal("Only root can extend a reservation longer than %v minutes", igorConfig.TimeLimit)
+			}
+		}
+
+		// Check to see if nodes are free to extend; if so, update the Schedule
+		for i := 0; i < duration; i++ {
```
jcrussell: This won't work if `MINUTES_PER_SLICE` isn't 1.

See https://github.com/sandia-minimega/minimega/blob/master/src/igor/schedule.go#L105

src/igor/extend.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Fatalln("cannot determine current user", err)
+	}
+
+	for _, r := range Reservations {
+		if r.ResName != subR {
+			continue
+		}
+
+		// The reservation name is unique if it exists
+		if r.Owner != user.Username {
```
jcrussell: Add `|| user.Username == "root"`?

csymonds:
LGTM

Issue #1016, opened 2017-11-02
Valid net driver detection


mkunz7:
Currently mini calls `kvm -device help` to determine what nics are compatible.

This isn't proper as these values change depending on the qemu binary and machine type

```
qemu-system-arm -M vexpress-a15 -net nic,model=?
<snip>
qemu: Supported NIC models: lan9118
```

mkunz7:
Doesn't seem to like any of the qemu, machine, cpu pairings I throw at it. 

pi3
```
vm config qemu /home/ubuntu/qemu-2.9.0/aarch64-softmmu/qemu-system-aarch64
vm config machine vexpress-a15
vm config cpu cortex-a53
Error (flax7): unable to determine valid QEMU CPUs -- exit status 1
```
```
/home/ubuntu/qemu-2.9.0/aarch64-softmmu/qemu-system-aarch64 -M vexpress-a15 -cpu help
Available CPUs:
<snip>
  cortex-a53
<snip>
```
pi2
```
vm config qemu /home/ubuntu/qemu-2.9.0/arm-softmmu/qemu-system-arm
vm config machine vexpress-a15
vm config cpu cortex-a15
```
pi1
```
vm config qemu /home/ubuntu/qemu-2.9.0/arm-softmmu/qemu-system-arm
vm config machine versatilepb
vm config cpu arm1176
```

log level debug
```
2017/11/07 10:57:50 DEBUG command_socket.go:180: got request over socket: {vm config cpu arm1176  }
2017/11/07 10:57:50 DEBUG external.go:259: cmd /home/ubuntu/qemu-2.9.0/arm-softmmu/qemu-system-arm completed in 10.069682ms
2017/11/07 10:57:50 DEBUG bridge.go:42: periodic reapTaps
```

mkunz7:
Can we add `vm config cpu` and `vm config machine` to the `vm config` printing?

jcrussell:
Can you send the full output of

`/home/ubuntu/qemu-2.9.0/aarch64-softmmu/qemu-system-aarch64 -cpu ?`

and

`/home/ubuntu/qemu-2.9.0/aarch64-softmmu/qemu-system-aarch64 -M vexpress-a15 -cpu ?`

mkunz7:
```
# /home/ubuntu/qemu-2.9.0/aarch64-softmmu/qemu-system-aarch64 -cpu ?
qemu-system-aarch64: No machine specified, and there is no default
[...]

Issue #911, opened 2017-05-03
Feature request: vm config smp


mkunz7:
```
vm config vcpus
```
changes the number of processors, but not the number of cores
```
vm config smp 4
vm config cores 2
```
should generate the qemu string
```
-smp 4,cores=2
```
which should give 8 total cores on 4 processors, if smp is set ignore vcpus

jcrussell:
This is easy to add. @djfritz: is it worth supporting?

PR #1070, opened 2018-02-20
Improve QEMU integration

### LINKED ISSUE #910: Feature request: vm config machine
similar to:
```
vm config cpu core2duo
```
Can we implement
```
vm config machine q35
```
Having to do this gets annoying
```
vm config qemu-override "-enable-kvm" "-machine q35 -enable-kvm"
```

jcrussell:
_WIP: Don't merge, yet._

Add `vm config machine` and `vm config cpu` based on output from the current `vm config qemu`. Make network drivers dependent on `vm config qemu` and `vm config machine`. Add `validate` and `suggest` to vmconfiger so that we can check the values for `vm config`s and suggest their values.

Also adds `vm config cores`.

Fixes #910, #911, and #1016.

@mkunz7


mkunz7:
:O :)

jcrussell:
Not sure the network driver detection is correct. `kvm -net nic,model=?` returns a different list than `kvm -device driver=?`. We should probably use the latter.

See note [here](https://wiki.qemu.org/Documentation/Networking#The_legacy_-net_option).

PR #1071, opened 2018-02-20
travis: bump go versions

jcrussell:
Build with latest 1.8 and 1.9 versions. Add 1.10.


jcrussell:
Hah, travis' regex seems broken for Go versions.

PR #1073, opened 2018-02-23
igor - convert schedule.json to schedule.gob 

csymonds:
Adding in a check and process to convert json to gob so we can make a clean transition

### DIFF
--- a/src/igor/main.go
+++ b/src/igor/main.go
@@ -347,8 +347,7 @@ func main() {
 	// Read in the schedule
 	path = filepath.Join(igorConfig.TFTPRoot, "/igor/schedule.gob")
 	usejson := false
-	_, err = os.Stat(path)
-	if err != nil {
+	if _, err = os.Stat(path); os.IsNotExist(err) {
 		log.Warn("failed to find gob schedule file: %v\nAttempting to open legacy json file...", err)
 		path = filepath.Join(igorConfig.TFTPRoot, "/igor/schedule.json")
 		scheddb, err = os.Open(path)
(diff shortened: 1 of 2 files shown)


jcrussell:
As we discussed -- let's make a `schedToGob` function that:

- Reads in `schedule.json`
- Writes to `schedule.gob`
- Deletes `schedule.json` if no errors

This should only run if there is a json file. If there is a gob file, igor should `log.Fatal`.

djfritz:
I don't see an posted issue in the tracker nor do I see any discussion on this. Where was the discussion?

I'm opposed to storing configs in non-human readable formats as a rule. The greater unix community is as well, barring crazy large configs that get a performance benefit from binary formats.

What's the compelling reason for this? @csymonds @jcrussell @floren 

Besides that - adding transition code in the codebase for the purpose of moving *2* known uses of this tool seems pretty nuts to me. If we do change the format, having the codebase migrate you is crazy.

csymonds:
This is indeed a case where the file gets crazy large and performance is a major issue we have run into on one of our clusters.

I agree that adding in code to do the transition is not ideal, in fact it is pretty ugly, however the concern is that someone out there in the world using this code might get blindsided with the change and we want to prevent that. I am certainly open to alternatives/suggestions.

jcrussell:
To echo what @csymonds said, because we store the node allocations for every minute from the current time until the end of the last reservation, the schedule file can get massive if we have long reservations.

We were seeing a 20 second delay just to decode the json. gob is much more efficient to decode and drops this delay to <1s. See #1053.

Ideally, we would change the format of the schedule file to be more sparse but that would require a lot more changes.

djfritz:
We change the minimega api all the time without providing a transition
tool, though I suspect I’ve just given @jcrussell a terrible idea...

How big are these files? Are we storing infinite history or something? What
is the measured performance difference here? I can’t imagine dozens or even
hundreds of reservations could generate so much data that we would be
hitting problems unless we’re just doing something wrong by storing
unnecessary data.

jcrussell:
We're going to implement this in a one-off tool instead.

floren:
If you change the MINUTES_PER_SLICE const in main.go to, say, 10, your reservations will have 10-minute granularity which is typically plenty for production clusters and causes a 10x reduction in reservation size.

PR #1078, opened 2018-03-08
Usb serial number

floren:
Allow the user to give a USB disk a serial number. Implemented this to track USB devices across a virtual enterprise.


jcrussell:
LGTM

PR #1080, opened 2018-03-13
minimega: add RESTful command interface

jcrussell:
_Please wait for 2.5 to merge._

POST JSON to :9006/command to run a minimega command. For example:
```
curl 'http://localhost:9006/command' -d '{"command": "vm info" }'
```
Should this be enabled by default or enabled via option?


djfritz:
How on parity is this with the current domain socket solution? If there's perfect overlap, we should consider deprecating the domain socket interface and updating all the things that use it. It's ugly to have to maintain so many interfaces.

jcrussell:
It should have parity with the exception of access control.

Another option would be to kill this and move it to `miniweb` which does JSON <-> GOB translation.

jcrussell:
We're going to implement `/command` in miniweb instead.

PR #1084, opened 2018-03-16
igor: usage statistics

csymonds:
This PR adds in a new 'stats' command that will scan the logs and pull basic usage statistics. Takes an argument for the number of days you want statistics for and has a verbose mode.


jcrussell:
Needs to support extended reservations now. #1062.

Issue #1095, opened 2018-04-19
`vm_net` test fails randomly


jcrussell:
```
## # disconnect single and multiple VMs				## # disconnect single and multiple VMs
## vm net disconnect k0 0					## vm net disconnect k0 0
## .columns name,vlan,bridge vm info				## .columns name,vlan,bridge vm info
name | vlan           | bridge					name | vlan           | bridge
c0   | [LAN3 (103)]   | [mega_bridge]				c0   | [LAN3 (103)]   | [mega_bridge]
c1   | [LAN3 (103)]   | [mega_bridge]				c1   | [LAN3 (103)]   | [mega_bridge]
k0   | [disconnected] | []					k0   | [disconnected] | []
k1   | [LAN3 (103)]   | [mega_bridge]				k1   | [LAN3 (103)]   | [mega_bridge]
## vm net disconnect c0 0					## vm net disconnect c0 0
E: remove port failed: no such port			      <
## .columns name,vlan,bridge vm info				## .columns name,vlan,bridge vm info
name | vlan           | bridge					name | vlan           | bridge
c0   | [LAN3 (103)]   | [mega_bridge]			      |	c0   | [disconnected] | []
c1   | [LAN3 (103)]   | [mega_bridge]				c1   | [LAN3 (103)]   | [mega_bridge]
k0   | [disconnected] | []					k0   | [disconnected] | []
k1   | [LAN3 (103)]   | [mega_bridge]				k1   | [LAN3 (103)]   | [mega_bridge]
## vm net disconnect all 0					## vm net disconnect all 0
E: remove port failed: no such port			      <
remove port failed: no such port			      <
## .columns name,vlan,bridge vm info				## .columns name,vlan,bridge vm info
```

I think adding a sleep after `vm start all` would fix the test but we should figure out why containers say they are "started" but the networking isn't setup.

PR #1097, opened 2018-04-24
minimega: add container-specific network functions

jcrussell:
Containers only have interfaces when they are running so check the state
before trying to interact with OVS. Otherwise, update the NetConfig and
return.


jcrussell:
Seems to reliably pass the `minitest` that was failing. #1095.

PR #1099, opened 2018-04-25
minimega: clear plumb in all

jcrussell:
Clear plumb state in clear all.

### DIFF
--- a/src/minimega/misc_cli.go
+++ b/src/minimega/misc_cli.go
@@ -283,9 +283,9 @@ func cliClearAll(c *minicli.Command, respChan chan<- minicli.Responses) {
 		"clear namespace all",
 		// clear vlan blacklist
 		"clear vlans all",
-		// clear pipes and plumbing
-		"clear pipe",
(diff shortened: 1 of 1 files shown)


src/minimega/misc_cli.go
```diff
@@ -283,6 +283,9 @@ func cliClearAll(c *minicli.Command, respChan chan<- minicli.Responses) {
 		"clear namespace all",
 		// clear vlan blacklist
 		"clear vlans all",
+		// clear pipes and plumbing
+		"clear pipe",
+		"clear plumb",
```
djfritz: These are backwards - you can't clear pipes with active plumbing, so we should clear plumb first.

djfritz: Clarification - you can't clear a pipe with an active writer, and any plumbing will show up as a writer on any involved pipes (except the first in a chain, which would be just a reader).

jcrussell: Incoming...

PR #1108, opened 2018-06-04
minimega/ron/miniccc: add ufs to miniccc

jcrussell:
Exposes the guest filesystem to the host over the cc connection.


jcrussell:
TODO: unmount if the VM is killed?

PR #1115, opened 2018-06-12
miniccc: look in files dir for executables

### LINKED ISSUE #1040: miniccc: smarter $PATH
If we push a binary, it would be nice to say `cc exec foo` rather than `cc exec /tmp/miniccc/files/foo`.

It would be pretty simple to add `/tmp/miniccc/files/` to `$PATH`.

jcrussell:
Check the files dir in addition to $PATH. Simplifies `cc exec`:

        cc send foo
        cc exec foo

Instead of:

        cc send foo
        cc exec /tmp/miniccc/files/foo

Fixes #1040

### DIFF
--- a/src/miniccc/commands.go
+++ b/src/miniccc/commands.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	log "minilog"
 	"os/exec"
+	"path/filepath"
 	"ron"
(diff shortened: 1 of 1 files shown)


jcrussell:
It checks ${PATH} first.

PR #1126, opened 2018-06-19
various: download files from miniweb

jcrussell:
Add `file stream` API to stream a file from meshage. Use in miniweb to download files.


jcrussell:
Fails minitest....

```
## file list /minitest                                          ## file list /minitest
dir   | name          | size                                    dir   | name          | size
<dir> | minitest/foo1 | 60                                    | <dir> | minitest/foo1 | 4096
<dir> | minitest/foo2 | 80                                    | <dir> | minitest/foo2 | 4096
<dir> | minitest/foo3 | 100                                   | <dir> | minitest/foo3 | 4096
```

Why do directories suddenly have a size!=4096?

mkunz7:
My understanding is directories contain a list of all their files within, by default they start at 4096, but grow as needed. 

```
mkdir test
cd test
for i in {1..1000}; do touch test$i; done
```

mkunz7:
Sorry getting over a cold, I see what you're saying now, that is weird. Gonna purge my comment.

PR #1127, opened 2018-06-20
minimega: bug fixes to `disk`

jcrussell:
`disk` should error when there is more than one partition. Fix cleanup.


jcrussell:
Need to fix minitest.

PR #1130, opened 2018-06-20
minimega: add `tar:` prefix

jcrussell:
Fetches and untars tarballs via meshage. Current only untars if there is
a single top-level directory. It untars to the same directory that
contains the tarball... we probably don't want that when the tarball is
in the iomBase. Maybe it would be better to always untar to
/tmp/minimega/???.

Thoughts? @djfritz


djfritz:
@jcrussell merged so let's see how this goes - my opinion is iomBase is the better place, but let's see what feedback we get.

PR #1132, opened 2018-06-22
igor: change time limits

jcrussell:
Time limits are now based on the number of nodes (well log of the number
of nodes). Extensions can be made as long as the reservation is near the
end (config-defined) and can be performed indefinitely.

### DIFF
--- a/src/igor/sub.go
+++ b/src/igor/sub.go
@@ -202,7 +202,13 @@ func runSub(cmd *Command, args []string) {
 		}
 	}
 	if user.Username != "root" {
-		if err := checkTimeLimit(len(nodes), duration); err != nil {
+		// nodes is only set if using subW
+		n := len(nodes)
+		if subN > 0 {
+			// must be using subN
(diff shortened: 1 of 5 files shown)


src/igor/sub.go
```diff
@@ -201,9 +201,9 @@ func runSub(cmd *Command, args []string) {
 			log.Fatal("Only root can make a reservation of more than %v nodes", igorConfig.NodeLimit)
 		}
 	}
-	if user.Username != "root" && igorConfig.TimeLimit > 0 {
-		if duration > igorConfig.TimeLimit {
-			log.Fatal("Only root can make a reservation longer than %v minutes", igorConfig.TimeLimit)
+	if user.Username != "root" {
+		if err := checkTimeLimit(len(nodes), duration); err != nil {
```
csymonds: If the user is using the -n flag, nodes will always be empty at this point. We need to use the value in subN in that case.

PR #1136, opened 2018-06-27
miniplumber: Added "message count" column to output from 'pipe' command.

mpopple:
Counts the total number of messages received on each pipe.

### DIFF
--- a/src/minimega/plumber.go
+++ b/src/minimega/plumber.go
@@ -188,7 +188,7 @@ func cliPipeBroadcast(ns *Namespace, c *minicli.Command, resp *minicli.Response)
 		}
 	} else {
 		// get info on all named pipes
-		resp.Header = []string{"name", "mode", "readers", "writers", "via", "last message"}
+		resp.Header = []string{"name", "mode", "readers", "writers", "via", "last message", "message count"}
 		resp.Tabular = [][]string{}
 
 		for _, v := range plumber.Pipes() {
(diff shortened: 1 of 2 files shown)


mpopple:
Good point, I agree; I'll come up with header names and make the change today.

tests/plumbing_miniccc.want
```diff
@@ -0,0 +1,29 @@
...
+
+## # Send a message from foo through foopipe
+## cc filter hostname=foo
+## cc exec stdout=foopipe echo "hello, world"
+
+## shell sleep 5s
+
+## # Verify that barpipe received the message from foopipe
+## pipe
+name    | mode | readers | writers | via | received     | count
```
jcrussell: Shouldn't this be `previous` not `received`?

src/minimega/plumber.go
```diff
@@ -188,7 +188,7 @@ func cliPipeBroadcast(ns *Namespace, c *minicli.Command, resp *minicli.Response)
 		}
 	} else {
 		// get info on all named pipes
-		resp.Header = []string{"name", "mode", "readers", "writers", "via", "last message"}
+		resp.Header = []string{"name", "mode", "readers", "writers", "via", "previous", "count"}
```
jcrussell: Let's change the order so that `count` comes after `writers`.

tests/plumbing_miniccc.want
mpopple: It should; sorry, I thought I had copied the .want changes over. Will fix.

src/minimega/plumber.go
mpopple: Will do.

PR #1137, opened 2018-06-28
igor: change tmp file write location

csymonds:
Changing the location of the tempdata file to TFTPRoot/igor/tmp data to ensure write permission


csymonds:
@jcrussell - modified

PR #1148, opened 2018-07-24
Add plumbing tests - vias and basic

HannahReinbolt:
Plumbing tests added for vias and basic plumbing. Also added is Jon's addition to the normal program for plumbing, which is the seed function.


jcrussell:
We're going tweak this and merge in pieces.

PR #1159, opened 2018-08-13
igor: show - list unsubbed nodes

csymonds:
adding a line in the igor show output listing the unreserved nodes.

### DIFF
--- a/src/igor/show.go
+++ b/src/igor/show.go
@@ -82,6 +82,8 @@ func runShow(_ *Command, _ []string) {
 
 	// Maps a node's index to a boolean value (up = true, down = false)
 	nodes := map[int]bool{}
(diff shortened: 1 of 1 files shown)


src/igor/show.go
```diff
@@ -130,8 +132,9 @@ func runShow(_ *Command, _ []string) {
 		nodes[v] = true
 	}
 
-	// Gather a list of which nodes are down
+	// Gather a list of which nodes are down and which nodes are unreserved
 	var downNodes []string
+	var unreservedNodes []string
 	for i := igorConfig.Start; i <= igorConfig.End; i++ {
```
jcrussell: You could move this code down and combine it with the loop that "compile[s] a list of unreserved nodes".

PR #1160, opened 2018-08-13
igor: show - sort by owner

csymonds:
Adding an option to the show command to allow sorting by owner name to make it easier to find reservations by owner

### DIFF
--- a/src/igor/show.go
+++ b/src/igor/show.go
@@ -59,16 +59,39 @@ const (
 )
 
 var cmdShow = &Command{
-	UsageLine: "show",
+	UsageLine: "show [-o]",
 	Short:     "show reservations",
 	Long: `
 List all extant reservations. Checks if a host is up by issuing a "ping"
+
+OPTIONAL FLAGS:
+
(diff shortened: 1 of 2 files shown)


src/igor/show.go
```diff
@@ -151,7 +174,11 @@ func runShow(_ *Command, _ []string) {
 	}
 	// nameFmt will create uniform color bars for 1st column
 	nameFmt := "%" + strconv.Itoa(maxResNameLength) + "v"
-	sort.Sort(StartSorter(resarray))
+	if subO {
+		sort.Sort(ByOwner(resarray))
```
jcrussell: Replace with https://golang.org/pkg/sort/#Slice

jcrussell:
Looks good. Do we want to sort each owner's reservations by start time?

```go
		sort.Slice(resarray, func(i, j int) bool {
			if resarray[i].Owner == resarray[j].Owner {
				return resarray[i].StartTime < resarray[j].StartTime
			}
			return resarray[i].Owner < resarray[j].Owner
		})
```

Issue #966, opened 2017-09-05
minimega: add `cc filter name=...`


jcrussell:
It would be nice to filter on VM name instead of setting `vm config tag name <name>`.

jcrussell:
We could add the `VM.Info` method to the `ron.VM` interface and filter on any `vm info` column...

PR #1161, opened 2018-08-20
minimega/ron: cc filter using vm info columns

jcrussell:
Make it so that you can use any "vm info" column in "cc filter".

Needs testing.


jcrussell:
Fixes #966.

PR #1167, opened 2018-08-22
minimega: fix capture

jcrussell:
Fixed issue with "unreachable".


src/minimega/misc.go
```diff
@@ -90,6 +90,15 @@ func (m *loggingMutex) Unlock() {
 	log.Info("unlocked: %v:%v", file, line)
 }
 
+// unreachable returns an error when we reach a condition that should be
```
djfritz: Holy cow

PR #1169, opened 2018-08-27
Igor sync

csymonds:
New igor command that will look for orphan reservation IDs in the Schedule data and attempt to purge them.

### DIFF
--- a/src/igor/sync.go
+++ b/src/igor/sync.go
@@ -21,12 +21,14 @@ OPTIONAL FLAGS:
 	`,
 }
 
+var subL bool // -l
 var subV bool // -v
 
 func init() {
 	// break init cycle
 	cmdSync.Run = runSync
 
+	cmdSync.Flag.BoolVar(&subL, "l", false, "")
 	cmdSync.Flag.BoolVar(&subV, "v", false, "")
 }
 
@@ -50,6 +52,23 @@ func runSync(cmd *Command, args []string) {
 			}
 		}
 	}
+	// we are only listing
+	if subL {
+		return
+	}
+
+	// purge the orphan IDs from the shedule
+	if len(IDs) > 0 {
+		if !subV {
+			fmt.Println("Purging Orphan IDs from Schedule...")
+		}
+		for _, oid := range IDs {
+			purgeFromSchedule(oid)
+		}
+		fmt.Println("Done.")
+		dirty = true
+	}
+
 }
 
 func getOrphanIDs() []uint64 {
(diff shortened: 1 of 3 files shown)


src/igor/sync.go
```diff
@@ -0,0 +1,111 @@
...
+var cmdSync = &Command{
+	UsageLine: "sync",
+	Short:     "synchronize igor data",
+	Long: `
+Does an internal check to verify the integrity of the data file.
+
+OPTIONAL FLAGS:
+
+	-v	Verbose - Prints additioanl information
+	`,
```
jcrussell: Add line for what -l does?

src/igor/sync.go
```diff
@@ -0,0 +1,111 @@
...
+	resIDs := make(map[uint64]bool)
+	// make a list of all reseration IDs that appear in the schedule
+	for _, s := range Schedule {
+		for _, n := range s.Nodes {
+			resIDs[n] = true
+		}
+	}
+	// Go through the reservations and turn off IDs we know about
+	for _, r := range Reservations {
+		resIDs[r.ID] = false
```
jcrussell: Instead you could `delete(resIDs, r.ID)` so that below you don't need to check the value of resIDs for the given `k` -- it should only be false.

src/igor/sync.go
```diff
@@ -13,27 +13,54 @@ var cmdSync = &Command{
...
 func init() {
 	// break init cycle
 	cmdSync.Run = runSync
 
-	cmdSync.Flag.BoolVar(&subL, "l", false, "")
-	cmdSync.Flag.BoolVar(&subV, "v", false, "")
+	cmdSync.Flag.BoolVar(&subF, "f", false, "")
+	cmdSync.Flag.BoolVar(&force, "force", false, "")
+	cmdSync.Flag.BoolVar(&subD, "d", false, "")
+	cmdSync.Flag.BoolVar(&dryRun, "dry_run", false, "")
```
jcrussell: Does this map the -dry-run? Has an underscore instead of hyphen.

src/igor/sync.go
```diff
@@ -0,0 +1,132 @@
...
+var quiet bool  // -quiet
+
+func init() {
+	// break init cycle
+	cmdSync.Run = runSync
+
+	cmdSync.Flag.BoolVar(&subF, "f", false, "")
+	cmdSync.Flag.BoolVar(&force, "force", false, "")
+	cmdSync.Flag.BoolVar(&subD, "d", false, "")
+	cmdSync.Flag.BoolVar(&dryRun, "dry_run", false, "")
```
jcrussell: Does this map to `-dry-run`? Seems like it would map to `-dry_run`.

src/igor/sync.go
csymonds: Weirdly, the hyphen was not working before, so I switched to underscore. But now it's working, so my problem must have been elsewhere. I'll switch back.

PR #1177, opened 2018-09-21
ron/miniccc/minimega: improving reconnect

jcrussell:
Add synchronization bytes to hopefully allow ron and miniccc to reconnect. Write the magic bytes in both directions to flush both sides.

Add redial to the things we do when we start a VM.

### DIFF
--- a/src/miniccc/main.go
+++ b/src/miniccc/main.go
@@ -8,6 +8,7 @@ import (
 	"encoding/gob"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	log "minilog"
 	"net"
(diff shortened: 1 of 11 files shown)


jcrussell:
Tested with a debian VM. If you shutdown the VM and then `vm start` it, the connection is re-established and you can issue new commands. If you restart the VM (using the OS restart mechanism), the connection is not re-established since minimega doesn't know to call redial. We could add a listener to fix this.

One potentially problematic thing is that the miniccc state is fresh so it doesn't know what the previous command was so it reruns them all.

mkunz7:
Can you have the agent generate a random number and save it to disk on first run? From then on out make this the agents unique identifier.

mkunz7:
Yeah, that's my intent. Perhaps we can add a persistence command. 
```
cc persistence true
```
If persistence is enabled it writes every command it executes to disk. On reconnect it phones home for the list of commands to run, and skips any it has done already.

djfritz:
Yea the original intent behind having command persist like they do currently is exactly that. If a vm cycles, it can re-ingest all of the past commands. 

This all looks good to me.

PR #1179, opened 2018-09-26
Vmbetter debootstrap args

jerry-cruz:
Enables passing arbitrary arguments from vmbetter to debootstrap, and includes a minimal ubuntu bionic vmbetter config example.


jcrussell:
Anything else to add or is this ready? I'll do a test to see if it works for me.

PR #1183, opened 2018-10-03
Igor notify conf

aherna:
Added a new field for the config file called expirationleadtime. It allows admins to set how much lead time in hours to notify users about pending reservation expiration . if no value is specified a default lead time of 24 hours is in effect

### DIFF
--- a/src/igor/config.go
+++ b/src/igor/config.go
@@ -78,7 +78,7 @@ type Config struct {
 	// Domain for email address
 	Domain string
 
-	//Igor Notify Notice time (in hours) is the amount of time before reservation expires users are notified
+	//Igor Notify Notice time (in minutes) is the amount of time before reservation expires users are notified
 	ExpirationLeadTime int
 }
 
--- a/src/igor/config.go
+++ b/src/igor/config.go
@@ -77,6 +77,9 @@ type Config struct {
 
 	// Domain for email address
 	Domain string
+
+	//Igor Notify Notice time (in hours) is the amount of time before reservation expires users are notified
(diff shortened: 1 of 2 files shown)


src/igor/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 
 	// Domain for email address
 	Domain string
+
+	//Igor Notify Notice time (in hours) is the amount of time before reservation expires users are notified
+	ExpirationLeadTime int
```
csymonds: Because all of our other duration values are in minutes, can you please convert this value to minutes for the sake of consistency?

csymonds: It looks like the notes still indicate that the value should be in hours. Can you please update?

jcrussell:
Probably should have commented on this sooner... don't we want the notifications to go out daily once we're within the window? I would think that an email to say you have 3, 2, 1 days left would be helpful and serve as a nice reminder if someone doesn't get to it when it's 3 days out.

aherna:
> 
> 
> Probably should have commented on this sooner... don't we want the notifications to go out daily once we're within the window? I would think that an email to say you have 3, 2, 1 days left would be helpful and serve as a nice reminder if someone doesn't get to it when it's 3 days out.

I can make a new issue request and work on it

csymonds:
I don't see anything wrong with a single notification, we do this as a courtesy. I suppose I could see an additional final notification, but I personally don't think daily is necessary. At the end of the day, my own preference is just one email at the time that extension becomes available.

PR #1184, opened 2018-10-03
minicli: speed up compilation

jcrussell:
Index patterns into a trie-like data structure and then compile commands by walking the trie.


jcrussell:
Found a bug, currently doesn't copy minicli flags to subcommands (e.g. preprocess):

```
namespace foo shell ls file:test
```

jcrussell:
And another:

```
$ namespace foo cc process kill all
2018/11/13 07:35:54 WARN trie.go:180: ambiguous command, found 2 possibilities
```

PR #1195, opened 2018-10-22
Igor admin restrictions

aherna:
Allows Admins to make extends without having the reservation to be in the extend window


src/igor/extend.go
```diff
@@ -57,10 +57,13 @@ func runExtend(cmd *Command, args []string) {
 	}
 
 	user, err := getUser()
+
```
csymonds: @jcrussell : I'm not trying to nit-pick, but is it typical to have an empty line between a call and checking the err from the call?

aherna: we added that line so we can test my no-ouo settings. do you want me to repush?

jcrussell: I don't put a newline.

aherna: changed

src/igor/extend.go
```diff
@@ -57,10 +57,13 @@ func runExtend(cmd *Command, args []string) {
...
 	if err != nil {
 		log.Fatalln("cannot determine current user", err)
 	}
 
+	log.Debug("Running Extend as %v", user.Username)
+
```
csymonds: @jcrussell This one?

src/igor/extend.go
jcrussell: @csymonds this one.

csymonds:
Having issues with getting the newline commit back in; starting a new branch and pr.

PR #1196, opened 2018-10-23
miniplumber: Fixed dropped message errors with plumbing minitests.

mpopple:
Added a sleep to basic tests and cleared plumbing after each test.

### DIFF
--- a/tests/epilog
+++ b/tests/epilog
@@ -1,1 +1,4 @@
 clear all
+
+clear plumb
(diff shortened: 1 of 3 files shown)


jcrussell:
Did you base your changes off master? It looks like something weird happened to the git history.

I would close this PR, make a new branch from master, and then redo the changes that you have in this branch (e.g. update epilog and add sleeps).

PR #1201, opened 2018-10-29
minitest: distributed tests for plumbing

mpopple:
Added distributed tests for VM-to-VM communications between hosts over miniplumber pipelines. Diagrammatically, this looks like:
 
               vm1 -> mm1 -> mm2 -> vm2
 
where the vm <-> mm communications are through cc and the mm->mm communications are through meshage.


jcrussell:
We should add a `cc exec wc -l pipe.out` to make sure that the messages reached the end of the pipe.

PR #1202, opened 2018-10-31
made changes so that root can powercylce any nodes for any reservation

aherna:


src/igor/power.go
```diff
@@ -126,7 +126,7 @@ func runPower(cmd *Command, args []string) {
 
 			resID := currentSched.Nodes[index-1]
 			for _, r := range Reservations {
-				if r.ID == resID && r.Owner == user.Username {
+				if r.ID == resID && r.Owner == user.Username || user.Username == "root" {
```
jcrussell: This should be `r.ID == resID && (r.Owner == user.Username || user.Username == "root")`, right?

https://play.golang.org/p/RTaMWHEb6IP

aherna: wouldnt it be equivalent because of commutative property going left to right?

T && F || T = T && (F||T)

aherna: the first check is a fast fail and the other two would both need to fail to through the error which is what we want

aherna: wrong thread^

PR #1203, opened 2018-11-01
miniweb: Graph Update

mwisely:
Update miniweb's graph page:

- Update to new d3
- Use a reactive framework (Vue.js)
- Add three graph options that the user can choose from on the fly
    - VLAN-Centric (nodes are VLANs; this what we had before)
    - Host-Centric (nodes are hosts connected to other hosts)
    - Dandelion (blue nodes are hosts, red are VLAN)
- Add a searchable list of VLANs to the page


aherna:
Closing because of other proposed changes

Issue #1217, opened 2018-11-18
minimega: `tap mirror <vm> <index> <vm> <index>`


jcrussell:
We currently support <tap> <tap> but it would be easy to look up tap names based on VM and interface. The only problem is that VMs need to be colocated for this command to work as is.

PR #1221, opened 2018-12-05
minimega: add vm->vm tap mirror

jcrussell:
Allows you to specify tap mirrors by VM names and interface indices.
Needs testing.

Updates #1217


jcrussell:
Should probably use `wrapVMTargetCLI` instead so that you can issue this command from the head node. It should return an error if the VMs are not colocated.

jcrussell:
Possible issues:

 - Interfaces for containers aren't created unless they are running.
 - If VM is in quit or error state, there might not be an interface for it.
 - When containers are restarted, we might need to re-mirror the interface.

Lots of corner cases with this API...

jcrussell:
@djfritz: thoughts on these possible issues? I'm leaning towards merging this as-is and seeing if users actually encounter these issues or not.

PR #1222, opened 2018-12-05
minimega: add `vm config colocate`

jcrussell:
Allow users to schedule VMs on the same host via `vm config colocate`.

 * Supports chaining -- a with b, b with c, c with d.
 * Cannot specify `vm config schedule` with `vm config colocate` to simplify the code.
 * Can specify `vm config schedule` on "root VM" and then colocate VMs with it.
 * Minor fix to `vmconfiger`.

Has some unit testing, need to do some more testing.


jcrussell:
Blocking #1224, #1225, #1226, #1227.

PR #1240, opened 2019-01-07
minimega: fix issues 1224, 1225, and 1226

jcrussell:
Add `vm config vga`, `vm config sockets`, and `vm config threads`.


jcrussell:
TODO: fix `vm_config_save` test.

PR #1260, opened 2019-02-28
igor: one refactor to rule them all

jcrussell:
Notable changes:

 * Remove schedule and compute host-availability on the fly
 * Remove data.gob and only use reservations.json
 * Replace StartTime and EndTime with Start and End which are time.Time
 * Replace Duration int with Duration time.Duration
 * Rename ResName to Name
 * Change power so that it errors if you specify a node you don't own
 * Collect globals into igor struct
 * Dedup some code by adding methods to Config
 * Move reservations-centric code into methods on new Reservations type
 * Remove sync schedule -- no schedule anymore

Probably full of bugs. Adds some unit tests (which pass), should add
more.

Also, changes ranges so that it errors if the values are out of range.

### DIFF
--- a/src/igor/backend.go
+++ b/src/igor/backend.go
@@ -14,23 +14,15 @@ type Backend interface {
 	Uninstall(*Reservation) error
 }
 
-func GetBackend() Backend {
-	if igorConfig.UseCobbler {
-		return NewCobblerBackend()
-	}
-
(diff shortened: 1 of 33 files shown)


jcrussell:
`Duration.Round` was added in Go 1.9. We should probably drop 1.8 support since it's really old.

PR #1284, opened 2019-03-25
minimega: rework disk snapshots, interfaces, and cache modes

jcrussell:
Remove the -snapshot flag from the qemu args and instead create
snapshots of the disks when the VM first launches. These are saved in
the VM instance directory and can be saved in the future.

TODO:

- [x] Fix other minitests
- [x] Reflow help to default width


djfritz:
What's the performance implication of doing this? Can we measure something like `vm launch kvm 500` with and without this PR? Otherwise LGTM.

jcrussell:
I'd like to understand that as well.

jasoncontrib:
This has implications for disk performance within the guest, too: previously, when using qemu's `-snapshot` flag, `unsafe` caching is always used for the disk. With this change, we should decide whether we want to also use `unsafe` caching on the disk snapshot we create, or have some other default cache behavior. We should also decide if we want different default cache behaviors for `vm config snapshot true` vs. `vm config snapshot false`

jasoncontrib:
@djfritz `vm launch kvm 500` is about 3% slower with this PR. The performance impact should be fairly minimal, since using `-snapshot` on qemu also creates a temporary snapshot disk anyway.

djfritz:
Sounds good @jasoncontrib. Reviewing now.

PR #1287, opened 2019-03-26
VM instance path symlinks

jasoncontrib:
Add symlinks to the VM instance directories.


djfritz:
I'll let @jcrussell merge...

PR #1307, opened 2019-04-09
igor: Add -json flag to 'igor show'

mwisely:
Add option for JSON encoded output to `igor show`

This helps igor web get a list of reservations without worrying about file locking.  Let igor handle that.


jcrussell:
Neat. Any reason for not sorting the reservations before json encoding them?

mwisely:
Nope, not at all. I'll move that sort call up.

PR #1326, opened 2019-06-05
minimega,miniweb: launch specific vm config, new VM page

jcrussell:
Add a page to miniweb to launch a saved VM config. Add option to specify VM config name with `vm launch`.


jcrussell:
Clicking from `/vms/new` to VM list sets the URL to `/vms/vms` which breaks the connect link.

PR #1329, opened 2019-06-19
miniplumber: Pipe truncate

mpopple:
***WIP: Don't merge yet

Modifies existing pipe API and adds "pipe truncate" to API.

The changes to the existing API are as follows (previous -> new):

	"pipe <pipe> <mode,> <all,round-robin,random>" -> "pipe name <pipe> <mode,> <all,round-robin,random>"
	"pipe <pipe> <log,> <true,false>" -> "pipe name <pipe> <log,> <true,false>"
	"pipe <pipe> <data>" -> "pipe send <pipe> <data>"
	"pipe <pipe> <via,> <command>..." -> "pipe send <pipe> <via,> <command>..."

The changes are meant to make some existing commands more intuitive. As a contrived example, previously, "pipe foo mode" -- which may not sound like a valid command -- would send a message "mode" to the pipe minimega//foo, while "pipe foo mode round-robin" would change the mode of minimega//foo to round-robin delivery. ("pipe foo 'mode round-robin'", meanwhile, would send a message "mode round-robin" to minimega//foo but leave the delivery mode unaltered.)

Additionally, these changes are designed to make the pipe API more extensible. In this spirit, the ability to truncate long messages in output to "pipe" has been introduced:

	"pipe <truncate,> <length>"

...where <length> is in number of characters to be printed.

Usage examples (previous -> new):

	pipe foo "a message" -> pipe send foo "a message"
	pipe foo mode round-robin -> pipe name foo mode round-robin
	pipe truncate 32

Documentation via minimega's "help" command has been updated to reflect these changes.

Some vendor code has also been gofmt'ed, but is otherwise unmodified.


jcrussell:
Seems reasonable. Do you have other ideas for "global" state for pipes?

Can you create a new PR that doesn't change all the vendor code though?

jcrussell:
See #1334

Issue #1262, opened 2019-03-05
ron/miniccc unable to send/fetch files/folders > 512kB over serial


mkunz7:
```
mkdir C:\logs
fsutil file createnew C:\logs\test100kb.txt 100000
fsutil file createnew C:\logs\test500kb.txt 500000
```

```
cc recv C:\\logs\\*.txt
```

If the logs folder is >512kB it will cause miniccc to only fetch a partial subset of the data from a partial subset of the cc clients. In my case 5/30 cc clients responded with <500kB data each. I haven't tested using ip instead of serial.

jcrussell:
We should implement file chunking like we do in `iomeshage`.

djfritz:
This is a duplicate of #1332. Closing.

jcrussell:
#1332 is the PR that would close this issue.

djfritz:
#1332 is merged, closing.

PR #1332, opened 2019-06-21
various: send cc files in chunks

jcrussell:
Instead of reading and sending the file in a single message, split it
across several. This allows you to send and recv files that exceed the
size of VM memory.

Seems to work for the updated test case, needs more testing though.

Updates #1262.


jcrussell:
`cc_io` test fails, needs triage.

djfritz:
@jcrussell status on this?

djfritz:
@jcrussell should we just merge and test on master since this is passing tests?

jcrussell:
I haven't had a chance to triage why the `cc_io` test was failing sometimes. Could merge with the expectation of having to triage a bit.

djfritz:
That's probably the best route...

PR #1337, opened 2019-07-02
minimega,minirouter: configure OSPF options

jcrussell:
Add capability to configure OSPF options and article describing tweaking
the costs of different interfaces.


djfritz:
Also we should have made better use of the 1337 id in this PR. Shameful.

jcrussell:
Missed opportunity. Guess we'll have to wait for #31337.

PR #1338, opened 2019-07-03
minimega: fix #1336

### LINKED ISSUE #1336: minimega: clogged containers
**Describe your environment**

1. minimega version: master
2. Linux distro/version: debian testing
3. Go compiler version: N/A
4. VM types: container

**Describe the bug**

Getting this error message on flush:

```ERROR vms.go:359: clogged vm 1: remove /tmp/minimega/namespaces/minimega/b03c6f48-71c5-4169-82c4-5749d0f8efa3: no such file or directory```

Possibly related to #1287.

**To Reproduce**

```
vm config filesystem /path/to/filesystem
vm launch container 1
vm kill all
vm flush
```

**Expected behavior**

No error message. Doesn't seem to be polluting filesystem though.

jasoncontrib:
create VM instance path symlinks for containers, too

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -831,7 +831,7 @@ func (vm *ContainerVM) launch() error {
 	// check, create a directory for it, and setup the FS.
 	if vm.State == VM_BUILDING {
 		if err := os.MkdirAll(vm.instancePath, os.FileMode(0700)); err != nil {
-			return fmt.Errorf("unable to create VM dir: %v", err)
+			return vm.setErrorf("unable to create VM dir: %v", err)
 		}
 
 		if vm.Snapshot {
@@ -840,6 +840,10 @@ func (vm *ContainerVM) launch() error {
 			}
 		} else {
 			vm.effectivePath = vm.FilesystemPath
+		}
+
+		if err := vm.createInstancePathAlias(); err != nil {
+			return vm.setErrorf("createInstancePathAlias: %v", err)
 		}
 	}
 
@@ -917,7 +921,7 @@ func (vm *ContainerVM) launch() error {
 
 		// create source directory if it doesn't exist
 		if err := os.MkdirAll(v, 0755); err != nil {
-			return err
+			return vm.setErrorf("start container: %v", err)
 		}
 	}
 	// denotes end of volumes
--- a/src/minimega/kvm.go
+++ b/src/minimega/kvm.go
@@ -642,7 +642,7 @@ func (vm *KvmVM) launch() error {
 	if vm.State == VM_BUILDING {
 		// create a directory for the VM at the instance path
 		if err := os.MkdirAll(vm.instancePath, os.FileMode(0700)); err != nil {
-			return fmt.Errorf("unable to create VM dir: %v", err)
+			return vm.setErrorf("unable to create VM dir: %v", err)
 		}
 
 		// Create a snapshot of each disk image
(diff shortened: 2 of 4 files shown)


jcrussell:
If the MkdirAll/Symlink fail, the VM will still be in the running state (and not error). It's probably fine, how often do those calls fail?

djfritz:
If you have the base directory in a special filesystem that doesn't allow symlinks (NFS, etc). There are lots of filesystems that this would go bonkers on.

djfritz:
... Now that said you probably wouldn't get this far because I bet any FS that doesn't support symlinks also won't allow a UDS, which mini tries to create early in main().

djfritz:
What if the symlink still existed from a prior broken run or something - shouldn't we bail?

jasoncontrib:
> If the MkdirAll/Symlink fail, the VM will still be in the running state (and not error). It's probably fine, how often do those calls fail?

If it fails, it should still return an error through the `errs` channel just like any of the other potential errors in that section? Do those also leave the VM in the `RUNNING` state?

> What if the symlink still existed from a prior broken run or something - shouldn't we bail?

Do you mean if a `namespaces\<namespace>` directory still exists from a prior run, we should bail? The subdirectories under that are UUIDs, so there shouldn't be collisions, but yeah we could bail if the `namespaces` folder exists at all, if that's the behavior we want.

djfritz:
nono - just if somehow the uuid or something already existed - which could happen if you were using `vm config uuid`

@jcrussell

jcrussell:
minimega doesn't allow you to launch two VMs in the same namespace with the same UUID.

jasoncontrib:
> What if the symlink still existed from a prior broken run or something - shouldn't we bail?

So this is addressed in the last commit now.

> If the MkdirAll/Symlink fail, the VM will still be in the running state (and not error). It's probably fine, how often do those calls fail?

@jcrussell Is this true? From my tests, if I introduce an error in MkdirAll/Symlink, it stays in `BUILDING` state, same as it would for other errors in that section?

jcrussell:
Err, yeah, I meant `BUILDING` not `RUNNING`.

I guess the question is, if the MkdirAll/Symlink fail, are we fine with an error being printed but the VM not being in the `ERROR` state.

jasoncontrib:
That applies not just to MkdirAll/Symlink, but also to writeVMConfig errors, right? Could we just set the vm state to `ERROR` if any of those fail?

jcrussell:
True, we might want to set the state if any of the three fail. Or, we could ignore it since the VM is functioning otherwise. Not sure.

@djfritz

djfritz:
I think it should head to `ERROR` - if we have scripting that depends on the symlink stuff outside of minimega, we should be able to catch errors at boot time. 

It sounds like it's hard to do that now? How does `Launch()` bail on other errors now?

jasoncontrib:
@djfritz sounds good, the latest commit has it go to "ERROR" state now.

jcrussell:
LGTM

djfritz:
Several things here are still broken somehow:

1. Some kind of race - when reading a .mm script, no containers get launched and instead I get "cgroups not ready". When I turn on debug logging, the same script works fine...

2. "quit" still takes several minutes to complete, even with just a few hundred containers on a single node. With 2.5-release, the same action is a few seconds.

jasoncontrib:
>     2. "quit" still takes several minutes to complete, even with just a few hundred containers on a single node. With 2.5-release, the same action is a few seconds.

@djfritz I don't believe the slowness was introduced by #1287, `quit` is slow on the parent commit b0e51e27 to that merge as well,

djfritz:
Hmm then we should find that regression and file an issue. It's taking 1 second per container to clean up an experiment. Having 8000 containers on a node and trying to `clear all` is painful...

jasoncontrib:
e28b548313016ec8cd12a889730106955760093f is the commit that introduced the slowness you're seeing. However, it's also the commit that fixed  `ERROR vms.go:357: clogged VM: overlay unmount: device or resource busy` upon quitting containers...

jasoncontrib:
@djfritz I'm not able to reproduce

> Some kind of race - when reading a .mm script, no containers get launched and instead I get "cgroups not ready". When I turn on debug logging, the same script works fine...

Can you send me the .mm script with which you're seeing that issue?

djfritz:
The "cgroups not ready" was a stuck node of mine, so disregard that. I'd like to test this with #1353 and then merge.

djfritz:
Tested with #1353 - seems to work. Quitting with 100 running containers only takes 1.34s.

@jcrussell @jasoncontrib any other hangups before merging?

jcrussell:
I think it's ready to merge.

PR #1346, opened 2019-07-09
minimega: shove cc command ID into Data

### LINKED ISSUE #1344: minimega: return cc command ID in Data
Should `cc exec`, `cc bg`, `cc send` ... echo the command ID associated with them so that scripts can used the ID when referencing the `cc commands` table?

@djfritz

jcrussell:
Makes it easier for scripts to figure out what command ID they issued.

Fixes #1344

### DIFF
--- a/src/minimega/cc_cli.go
+++ b/src/minimega/cc_cli.go
@@ -369,7 +369,7 @@ func cliCCFileSend(ns *Namespace, c *minicli.Command, resp *minicli.Response) er
 
 	}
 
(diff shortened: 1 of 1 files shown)


jcrussell:
Surprised that you hate this; it seemed like a good way to return the information without echo'ing the command ID for every command. Maybe there's a better way to know what command ID that you're on without hardcoding based on number issued so far or, as I've seen done, adding one to the last ID in `cc commands`.

PR #1347, opened 2019-07-11
minimega: warn if 4+ disks

### LINKED ISSUE #1320: minimega: warn if more that 4 ide devices
CDROM is always `ide0` so you can only have 3 disks using `ide`. Warn if there are 4+ disks and the disk drivers are all `ide`.

jcrussell:
Fixes #1320


jasoncontrib:
@djfritz Moving off IDE, I'd be worried about breaking out-of-box compatibility with older operating systems.

djfritz:
Sure but that still leaves the problem of how you shovel coal into the VM to power it...

PR #1350, opened 2019-07-11
minimega: shorten the sleep between container filesystem unmount retries

jasoncontrib:
e28b548 introduced retries of filesystem unmounts for container flushes, but the sleep in between retries was 1 second. It's shortened in this PR.

### DIFF
--- a/src/minimega/container.go
+++ b/src/minimega/container.go
@@ -1266,6 +1266,8 @@ func (vm *ContainerVM) overlayMount() error {
 }
 
 func (vm *ContainerVM) overlayUnmount() error {
+	sleepAmount := time.Duration(1)
+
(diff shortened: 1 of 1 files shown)


djfritz:
And all of this is serialized somewhere? It looks like no matter what we hit the first retry from my debug logging, so we certainly are trying to unmount too quickly after killing the container.

jasoncontrib:
@djfritz Yeah, that's what I see, too. There's always the first retry.

djfritz:
@jcrussell what's the reason we serialize flush again?

jcrussell:
We could parallelize it, should work.

djfritz:
Collapsing this into #1353 as that may be a better fix.

PR #1351, opened 2019-07-11
vendor: update gopacket, net, sys

jcrussell:
Add go1.12.


djfritz:
... really gaming the LOC count for you eh?

jcrussell:
Hah, if we want to use go v1.12, we need to update gopacket. If we update gopacket, we have to add x/sys and x/net/bpf. If we add those, we have to give minimega a glass of milk.

Issue #1290, opened 2019-03-27
minimega: add `vm net add/remove`


jcrussell:
Extend the `vm net` API to add/remove interfaces using QMP commands.

djfritz:
Why do we need this? Hotplugging PCI devices can do all sorts of weird things, especially to windows VMs. Do we need this for 2.6? 

@jcrussell

djfritz:
Roger - do we need this for 2.6 then?

csymonds:
Yes. My goal is to have this one done within a week.

mkunz7:
Might be worth looking into hotplugging usb-net devices.

djfritz:
Kicking this to 2.7

csymonds:
Note: net add will be part of 2.7 release, net remove likely in 2.8

PR #1363, opened 2019-08-13
minimega: vm net add

csymonds:
This pr represents progress towards satisfying #1290 

Adding vm net add
Allows the user to hotplug a nic to a kvm, and specify some attributes - bridge, vlan, mac, and driver
Does not work for containers
Does not include the remove feature


src/minimega/vm_cli.go
```diff
@@ -834,7 +853,20 @@ func cliVMNetMod(ns *Namespace, c *minicli.Command, resp *minicli.Response) erro
...
-		if c.BoolArgs["disconnect"] {
+		if c.BoolArgs["add"] {
+			// This will do the work of adding the interface to the vm
+			nics, err := ns.parseVMNets(c.ListArgs["netspec"])
+			if err != nil {
+				return true, err
+			}
+			kvm, ok := vm.(*KvmVM)
+			if !ok {
+				return true, fmt.Errorf("Unable to get Kvm")
```
aherna, line 864: Maybe we rephrase to be less cryptic like "target vm is not a KVM"

tests/vm_add_nic
```diff
@@ -0,0 +1,84 @@
...
+vm launch kvm bar
+vm start all
+
+# wait for cc to connect
+shell sleep 30
+
+.column name,vlan,bridge,ip vm info
+
+# try adding a nic to foo on the same vlan as bar
+vm net add foo 100
```
aherna, line 18: Probably should add line to test the full command, vm net add <target> <bridge,vlan,mac,driver>
"vm net add foo test,100,3c:d9:ce:82:00:88"

make sure both vm nics are on same bridge

tests/vm_add_nic.want
```diff
@@ -0,0 +1,107 @@
...
+## vm start all
+
+## # wait for cc to connect
+## shell sleep 30
+
+## .column name,vlan,bridge,ip vm info
+name | vlan  | bridge        | ip
+bar  | [100] | [mega_bridge] | []
+foo  | []    | []            | []
+
```
aherna, line 19: If there are changes above with full command this output will be different

aherna (commented):
LGTM, minor revision to make sure error code is less cryptic, and test actually tests full command.

aherna (changes requested):
Need to resolve conflict namespace.go, change error message to be more obvious, change test to test full command

aherna (approved):
LGTM added new issue #1438

PR #1369, opened 2019-09-26
Add -vlan flag to sub and edit

mwisely:
Allows users to specify which VLAN they'd like a reservation to connect to. The value of `-vlan` can be:

- the name of another reservation (`string`)
    - The specified reservation must be writable by the current user
    -  The new reservation uses the same VLAN as the existing one
-  a VLAN ID (`int`)
    - ... As long as the specified VLAN ID is not in use by any other reservation, or the VLAN ID is used by at least one existing reservation that is writable for the current user.
    - The VLAN ID must be in the range specified by igor's config

If `-vlan` is not provided, igor determines a VLAN ID as usual.

`-vlan` can be specified for `sub` and `edit`

### DIFF
--- a/src/igor/reservations.go
+++ b/src/igor/reservations.go
@@ -93,11 +93,14 @@ func (r *Reservations) Install(res *Reservation) error {
 		return nil
 	}
 
-	// pick a network segment
-	if v, err := r.NextVLAN(); err != nil {
-		return fmt.Errorf("error setting network isolation: %v", err)
-	} else {
-		res.Vlan = v
+	// Vlan wasn't specified by flag
+	if res.Vlan == 0 {
+		// pick a network segment
+		if v, err := r.NextVLAN(); err != nil {
+			return fmt.Errorf("error setting network isolation: %v", err)
+		} else {
+			res.Vlan = v
+		}
 	}
 
 	// update network config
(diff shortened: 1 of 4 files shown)


src/igor/reservations.go
```diff
@@ -93,11 +93,14 @@ func (r *Reservations) Install(res *Reservation) error {
 		return nil
 	}
 
-	// pick a network segment
-	if v, err := r.NextVLAN(); err != nil {
-		return fmt.Errorf("error setting network isolation: %v", err)
-	} else {
-		res.Vlan = v
+	// Vlan wasn't specified by flag
+	if res.Vlan == 0 {
```
mwisely: If anyone has opinions about this line, I'd appreciate it. There may be a better way to indicate that igor needs to pick a VLAN.

csymonds: Where is this value getting set? I'm not seeing it...

mwisely: Go sets it to 0 by default, since it's an `int`. I assumed that if the `Vlan` field was 0, then the user must not have specified a VLAN with `-vlan`.  

I assumed that checking if it was zero would be OK because igor's configured VLANMin would probably always be greater than zero. That, and since VLAN zero is special, no one would ever try to choose it with `-vlan`.

csymonds: Gotcha. Yeah, given the circumstances, I don't think a little in-band signaling here will do us harm, though I share your discomfort. Given that igor is set to be overhauled, I don't think it is worth dwelling on the issue at this time. Let's move forward.

PR #1382, opened 2019-10-28
miniweb: Update packages to latest versions

mpopple:
Packages include:
jQuery (3.1.1 -> 3.4.1)
dataTables (1.10.12 -> 1.10.20)
bootstrap (3.3.5 -> 4.3.1)


djfritz:
@mwisely can you comment? There's no clean way to test this so I think we just need a sanity check re: versioning etc. Nobody knows this stuff on our team better than you.

mwisely:
@mpopple @djfritz It seems to work for me, so far. The listed versions appear to be the latest stable releases, and my JS console is clear.

I'm curious if we'll see any weirdness from updating Bootstrap 3.x to 4.x, but we can handle that if/when it arises.

Looks good to me, though!

mpopple:
Thanks @djfritz and @mwisely !

Issue #1392, opened 2020-01-02
minimega: specify `flush` target


glattercj:
Need ability to specify specific VM target(s) on `vm flush`. Backwards compatibility needs to be maintained (still need to be able to run without target args).

e.g.

```
"vm <flush,>",
"vm <flush,> <vm target>",
```

PR #1393, opened 2020-01-02
minimega: specify `flush` target

glattercj:
- Changed 'all' VM flush method to 'FlushAll()'.
- Created new 'Flush()' method that uses the 'Apply' method to flush a specified target.
- Updated help strings to reflect enhancement.
- Backwards compatible with old `vm flush` behavior

Resolves https://github.com/sandia-minimega/minimega/issues/1392

### DIFF
--- a/src/minimega/vms.go
+++ b/src/minimega/vms.go
@@ -343,6 +343,8 @@ func (vms *VMs) Kill(target string) error {
 
 // Flush VMs matching target.
 func (vms *VMs) Flush(target string, cc *ron.Server) error {
+	var mapLock sync.Mutex
+
 	return vms.Apply(target, func(vm VM, _ bool) (bool, error) {
 		if vm.GetState()&(VM_QUIT|VM_ERROR) == 0 {
 			return false, nil
@@ -352,13 +354,16 @@ func (vms *VMs) Flush(target string, cc *ron.Server) error {
 
 		if err := vm.Disconnect(cc); err != nil {
 			log.Error("unable to disconnect to cc for vm %v: %v", vm.GetID(), err)
-			return false, err
+			return true, err
 		}
 
 		if err := vm.Flush(); err != nil {
 			log.Error("clogged vm %v: %v", vm.GetID(), err)
-			return false, err
+			return true, err
 		}
+
+		mapLock.Lock()
+		defer mapLock.Unlock()
 
 		delete(vms.m, vm.GetID())
 
--- a/src/minimega/vms.go
+++ b/src/minimega/vms.go
@@ -341,9 +341,34 @@ func (vms *VMs) Kill(target string) error {
 	})
 }
 
-// Flush deletes VMs that are in the QUIT or ERROR state, disconnecting them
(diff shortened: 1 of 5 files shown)


src/minimega/vms.go
```diff
@@ -341,9 +341,34 @@ func (vms *VMs) Kill(target string) error {
...
+			log.Error("unable to disconnect to cc for vm %v: %v", vm.GetID(), err)
+			return false, err
+		}
+
+		if err := vm.Flush(); err != nil {
+			log.Error("clogged vm %v: %v", vm.GetID(), err)
+			return false, err
+		}
+
+		delete(vms.m, vm.GetID())
```
jcrussell: This needs to be synchronized otherwise it will panic.

activeshadow: @glattercj to elaborate on this more, `vms.Apply` calls the provided function once per matching target VM found, each in their own Goroutine. As such, similar to how things are done in the (what's now called) `FlushAll` function below, the call to `delete` VMs from the map needs to be synchronized with a mutex lock.

So, for example:

```
func (vms *VMs) Flush(target string, cc *ron.Server) error {
  var mapLock sync.Mutex

  return vms.Apply(target, func(vm VM, _ bool) (bool, error) {
    ...

    mapLock.Lock()
    defer mapLock.Unlock()

    delete(vms.m, vm.GetID())

    return true, nil
  }
}
```

src/minimega/vms.go
```diff
@@ -341,9 +341,34 @@ func (vms *VMs) Kill(target string) error {
...
+	return vms.Apply(target, func(vm VM, _ bool) (bool, error) {
+		if vm.GetState()&(VM_QUIT|VM_ERROR) == 0 {
+			return false, nil
+		}
+
+		log.Info("deleting VM: %v", vm.GetID())
+
+		if err := vm.Disconnect(cc); err != nil {
+			log.Error("unable to disconnect to cc for vm %v: %v", vm.GetID(), err)
+			return false, err
```
activeshadow, line 355: @glattercj based on the documentation for `vms.Apply`, you may want to `return true, err` here.

src/minimega/vms.go
```diff
@@ -341,9 +341,34 @@ func (vms *VMs) Kill(target string) error {
...
+		}
+
+		if err := vm.Flush(); err != nil {
+			log.Error("clogged vm %v: %v", vm.GetID(), err)
+			return false, err
```
activeshadow, line 360: @glattercj see previous comment.

glattercj: Fixed in latest commit

src/minimega/vms.go
glattercj: Fixed in latest commit

src/minimega/vms.go
glattercj: Fixed in latest commit

glattercj:
@csymonds @aherna Can you guys review this por favor?

glattercj:
@aherna bump

Issue #1413, opened 2020-04-30
minimega: allow "minimega" group to execute commands


ccordi:
We would like users to be able to run minimega commands without needing to be root. Instead, it would be preferable if all users that are part of the "minimega" group could execute commands.

This functionality could be achieved by running something similar to the following after starting minimega:

```
  chgrp -R minimega $MM_RUN_PATH
  chmod -R g=u $MM_RUN_PATH
  chmod g+s $MM_RUN_PATH ${MM_RUN_PATH}/minimega
```

This could likely be done in the systemctl script or by minimega on startup.

PR #1414, opened 2020-05-11
minimega: create minimega group in deb package and use it in init script

ccordi:
Resolves https://github.com/sandia-minimega/minimega/issues/1413


csymonds (approved):
LGTM

PR #1416, opened 2020-05-20
update to so git treats build artifacts of uminirouter as ignored

afbagwell:


afbagwell:
Retracting this because I think I really screwed up my fork. starting over

PR #1417, opened 2020-05-20
Update so git treats build artifacts of uminirouter as ignored.

afbagwell:
Changed perms on build.bash files from 664 to 775 so users don't get a git modified flag upon changing it themselves.


aherna (approved):
LGTM

PR #1419, opened 2020-06-08
minimega.namespace_cli: don't use loopback addresses for GRE tunnels

ccordi:
This prohibits loopback addresses from being used as endpoints in bridges in cliNamespaceBridge in minimega.namespace_cli.go.


csymonds (approved):
LGTM

Issue #1420, opened 2020-06-19
API page is missing


mzy2240:
All the web docus related to API are missing.

mkunz7:
There's a copy on https://tip.minimega.org/articles/api.article

PR #1423, opened 2020-07-09
appengine: restore api documentation on minimega.org

csymonds:
Modified the update script to remove existing update files which will cause the minimega build to fail when the update is run. 
Also included automation for pulling in the google cloud sdk. Tested and worked locally, but was not written with other dev environments in mind and could fail. For example, the google cloud sdk being pulled in is 64-bit x86.
After running this, I have confirmed that the API Documentation link on minimega.org has been restored. This resolves issue #1420


aherna (approved):
LGTM

PR #1424, opened 2020-07-12
Check for existance of node disk image and mark node as DNB if image is missing.

activeshadow:


activeshadow:
Crap... wrong source for pull request. Disregard!

PR #1432, opened 2020-10-23
igor: extend clarification

csymonds:
Clarifying logs for extend when duration exceeds limits. Before, the message indicated that the user request more time than specified in their command because we were adding it to the time remaining in their reservation.

Now, we specify that the requested time, plus the time remaining, is what exceeds the limit.


afbagwell (approved):
LGTM

PR #1436, opened 2020-12-17
miniccc: attempt to reconnect to serial cc on disconnect

activeshadow:
`miniccc` clients currently fail to reconnect if the `miniccc` process is restarted (for example, when a VM is rebooted). This PR addresses the issue.

### DIFF
--- a/src/ron/server.go
+++ b/src/ron/server.go
@@ -799,9 +799,22 @@ func (s *Server) clientHandler(c *client) {
 		}
 	}
 
-	if err != io.EOF && !strings.Contains(err.Error(), "connection reset by peer") {
-		log.Errorln(err)
+	// This is an OK error - likely just means the TCP connection was closed.
+	if err == io.EOF {
+		return
 	}
+
+	// This is an OK error - likely just means the client closed the connection.
+	if strings.Contains(err.Error(), "connection reset by peer") {
+		return
+	}
+
+	// This is an OK error - likely just means the unix socket was closed.
+	if strings.Contains(err.Error(), "use of closed network connection") {
+		return
+	}
+
+	log.Errorln(err)
 }
 
 // addClient adds a client to the list of active clients
--- a/src/ron/server.go
+++ b/src/ron/server.go
@@ -806,11 +806,13 @@ func (s *Server) clientHandler(c *client) {
 
 	// This is an OK error - likely just means the client closed the connection.
 	if strings.Contains(err.Error(), "connection reset by peer") {
+		log.Debugln(err)
 		return
 	}
 
 	// This is an OK error - likely just means the unix socket was closed.
 	if strings.Contains(err.Error(), "use of closed network connection") {
+		log.Debugln(err)
 		return
 	}
 
--- a/src/ron/server.go
+++ b/src/ron/server.go
@@ -203,43 +203,76 @@ func (s *Server) ListenUnix(path string) error {
 
 // Dial a client serial port. The server will maintain this connection until a
 // client connects and then disconnects.
-func (s *Server) DialSerial(path string) error {
-	log.Info("dial serial: %v", path)
+func (s *Server) DialSerial(path string, ccPortOpened, ccPortClosed chan struct{}) error {
+	dial := func(path string) (net.Conn, error) {
+		log.Info("dial serial: %v", path)
 
-	s.connsLock.Lock()
-	defer s.connsLock.Unlock()
-
-	// are we already connected to this client?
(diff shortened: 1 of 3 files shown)


activeshadow:
FYI, just tested this with Windows and it doesn't work (I had only tested with Linux originally). Seems the custom `virtioPort io.ReadWriteCloser` for Windows doesn't return an EOF like the Linux implementation does.

activeshadow:
This now works for both Windows and Linux VMs.

aherna (commented):
LGTM some observations when issuing commands on both win10 and ubuntu18 and the machines restart the connections are closed but since the commands are still in queue it attempts to issue the commands resulting in following error: "read unix @->/tmp/minimega/0/cc: use of closed network connection"
when machines comes back the error goes away. can you add a patch so that it doesnt attempt use a closed connection?

activeshadow:
> @activeshadow LGTM some observations when issuing commands on both win10 and ubuntu18 and the machines restart the connections are closed but since the commands are still in queue it attempts to issue the commands resulting in following error: "read unix @->/tmp/minimega/0/cc: use of closed network connection"
> when machines comes back the error goes away. can you add a patch so that it doesn't attempt use a closed connection?

@aherna I assume it's the ron server printing this error message, and not the miniccc client? Can you provide enough detail so I can reproduce the error message in order to make sure I address it properly?

activeshadow:
@aherna can you confirm it's the ron server printing this error message?

Also, do the commands being sent when this error is printed get lost, or do they get queued up again since they weren't actually sent?

activeshadow:
@aherna I was able to reproduce and answer my questions above.

aherna:
> @aherna can you confirm it's the ron server printing this error message?
> 
> Also, do the commands being sent when this error is printed get lost, or do they get queued up again since they weren't actually sent?

@activeshadow Im 90% sure that its server.go in src/ron. The only other place that shows up is in conatiners.go,  minitunnel.go, and kvm.go. src/minimega/kvm.go would be my next logical thought of where it was throwing this error.

The commands stay queued they dont get lost, unless user issues clear cc commands.

activeshadow:
@aherna I don't think this error is hurting anything, since commands don't get lost.

The error is being returned by the call to `c.dec.Decode` in `ron/server.go` [here](https://github.com/activeshadow/minimega/blob/37b39452ec606c1d008e852d54a3dbeb39765927/src/ron/server.go#L761). `c.dec` is a Gob decoder tied to the unix socket for the client's virtual serial port, so it returns this error when the unix socket is closed when we detect the `VSERPORT_CHANGE` event when the VM reboots.

Looking at my logs, and based on how the code was before I added the `VSERPORT_CHANGE` stuff, it looks like this was happening even before I made the changes in this PR.

There's a few ways to fix it... one is to do what's currently being done at the end of the `clientHandler` function now and just not print the error if it's this `read unix` error. It's a common approach to block on read and handle these types of errors like this. The other would be to have another for loop that checks to see if the connection is closed or if data is ready via something like `bufio.Peek` before moving on to `c.dec.Decode` but this seems unnecessary.

Thoughts? Would be great to get this merged in!

src/ron/server.go
```diff
@@ -766,9 +799,22 @@ func (s *Server) clientHandler(c *client) {
...
-	if err != io.EOF && !strings.Contains(err.Error(), "connection reset by peer") {
-		log.Errorln(err)
+	// This is an OK error - likely just means the TCP connection was closed.
+	if err == io.EOF {
+		return
 	}
+
+	// This is an OK error - likely just means the client closed the connection.
+	if strings.Contains(err.Error(), "connection reset by peer") {
+		return
```
aherna, line 809: Lets log this at least as info or debug and we can approve

src/ron/server.go
```diff
@@ -766,9 +799,22 @@ func (s *Server) clientHandler(c *client) {
...
+	}
+
+	// This is an OK error - likely just means the unix socket was closed.
+	if strings.Contains(err.Error(), "use of closed network connection") {
+		return
```
aherna, line 814: Lets log this at least as info or debug and we can approve

aherna (changes requested):
Lets log the closed connections and call it a day for this PR

src/ron/server.go
activeshadow: Debug log added. :+1:

src/ron/server.go
activeshadow: Debug log added. :+1:

aherna (approved):
LGTM

PR #1437, opened 2020-12-22
minimega: use "actual path" from qemu-img for backing image if present

activeshadow:
If you run `qemu-img info` in the same directory as the image your passing to it, then the `(actual path: ...)` stuff isn't included in the output. However, if you run `qemu-img info` from a different directory, or you pass the absolute path of the image to `qemu-img info`, then the `(actual path: ...)` stuff is present. Minimega always passes the full path of a disk image to `qemu-img info`.

As an aside, I'm not sure if this is new functionality in a more recent version of `qemu-img` (I have the following version of `qemu-utils` installed: `focal-updates,focal-security,now 1:4.2-3ubuntu6.10`).

### DIFF
--- a/src/minimega/disk.go
+++ b/src/minimega/disk.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


aherna (approved):
LGTM

PR #1429, opened 2020-09-10
miniweb: minibuilder

aherna:

PR #1439, opened 2021-01-08
minimega: Resolve gofmt issue from #1429

csymonds:
Trivial change, but across many files: gofmt failed for non-obvious reasons. I ran gofmt and wrote changes back to the files to resolve and now we appear to build correctly and pass gofmt.


aherna (approved):
LGTM

PR #1441, opened 2021-01-11
doc: Download page for minimega.org

csymonds:
Adding a downloads page. 
Minor fix on link menu. 
Removed the version file which was added previously in error.


aherna (approved):
LGTM

Issue #1431, opened 2020-10-09
Can't build some components


floren:
**Describe your environment**

1. master (75fe6c97a83e143bd61ca5e0744c12f05c16e501)
2. Debian, kernel 4.19.0
3. Go 1.15.2 linux/amd64
4. N/A

**Describe the bug**

I want to update my binaries so I can use a more recent protonuke. build.bash complains about a dependency:

```
john@thufir:~/minimega$ ./build.bash 
BUILD PACKAGES (linux)
apigen
bridge
gonetflow
igor
igorweb
iomeshage
meshage
miniccc
minicli
miniclient
minidoc
minifuzzer
minilog
minimega
minipager
miniplumber
minirouter
minitest
minitunnel
miniweb
nbd
nfcat
passwordify
src/passwordify/main.go:10:2: code in directory /home/john/minimega/src/vendor/golang.org/x/crypto/ssh/terminal expects import "ssh/terminal"
```

and so does trying to build directly:

```
john@thufir:~/minimega/src/protonuke$ GOPATH=/home/john/minimega/ go build
ssh.go:18:2: code in directory /home/john/minimega/src/vendor/golang.org/x/crypto/ssh expects import "ssh"
ssh.go:19:2: code in directory /home/john/minimega/src/vendor/golang.org/x/crypto/ssh/terminal expects import "ssh/terminal"
```

What's the right way to build things?

mkunz7:
use go 1.12 for now ku.nz/miniclass

jecompton:
Actually, this is just the go format checker dying on comments. In `src/vendor/golang.org/x/crypto/ssh/terminal/util.go` and `src/vendor/golang.org/x/crypto/ssh/doc.go` there are comments on the package line like `package terminal // import "ssh/terminal"`.

Delete the double-slash to the end of the line for those two lines and you should be able to build.

jecompton:
> Actually, this is just the go format checker dying on comments. In `src/vendor/golang.org/x/crypto/ssh/terminal/util.go` and `src/vendor/golang.org/x/crypto/ssh/doc.go` there are comments on the package line like `package terminal // import "ssh/terminal"`.
> 
> Delete the double-slash to the end of the line for those two lines and you should be able to build.

Oops, maybe it's not just a format error. I think it's actually trying to parse it as something like `package terminal//import` and thinks it is trying to include itself. 

Anyway, the fix is the same, and the lesson is don't put comments on `package` lines in .go code.

[...]

PR #1442, opened 2021-01-14
protonuke: Fixed go vet issue

aherna:
Fixed the vetting issue that is caused with newer versions of Go


aherna:
Closes #1431

csymonds (approved):
LGTM
Note: I have tested compilation with go versions 1.13.15, 1.14.13, and 1.15.6 and had no issues.

PR #1450, opened 2021-04-06
minimega: Update install docs on minimega.org

csymonds:
Updating installation documentation on minimega site to clarify dependencies
Reorganizing and clarifying dependencies based on suggestions


jacdavi:
Closing since we've moved to sandia.gov/minimega. Looking through the changes, I don'r see anything obvious that's missing from the current website.

Issue #1445, opened 2021-02-16
Bug: Unpaused vms cause migration issues


mkunz7:
**Describe your environment**

1. minimega version - latest
2. Linux distro/version - ubuntu 20.04
3. Go compiler version - 1.12.5
4. VM types - kvm

**Describe the bug**
if you don't pause a vm with `vm stop` before you create a migration file with `vm migrate`, `vm migrate` will never finish and show completion statuses >100%. I stopped mine at 300%.

**To Reproduce**
Tested with a windows 10 image.
```
clear vm config
vm config memory 4096
vm config disk /root/windows.qcow2
vm config experiment,main
vm launch kvm windows
vm start windows
vm migrate windows /root/windows.state
vm migrate
vm migrate
vm migrate
```
vs
```
clear vm config
vm config memory 4096
vm config disk /root/windows.qcow2
vm config experiment,main
vm launch kvm windows
vm start windows
vm stop windows
vm migrate windows /root/windows.state
vm migrate
```
**Expected behavior**
`vm migrate` should either warn the vm is not stopped or stop it for you

GhostofGoes:
closing since this was probably fixed

PR #1451, opened 2021-05-17
minimega: Enhance ns snapshot

### LINKED ISSUE #1309: minimega: add disk support to `ns snapshot`
With #1284 and #1302, it should be possible to create snapshots of disk images as well. This could allow the whole experiment state to be bundled up and saved.

glattercj:
Closes #1309 

- Adds disk state save to `ns snapshot` command
- For each VM:
  - First the disk state is saved using the `drive-backup` qmp command (__*BLOCKS until finished; later plans to implement goroutine for this*__)
  - Next the rest of the VM state is saved using the normal `migrate`
- Addresses issue in #1445 by explicitly pausing all VMs as before proceeding with the snapshot

### DIFF
--- a/doc/content/articles/namespaces.article
+++ b/doc/content/articles/namespaces.article
@@ -287,14 +287,36 @@ namespace.
 
 ** Snapshot
 
-Namespaces are working towards a full snapshot of a running experiment. This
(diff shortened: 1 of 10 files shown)


csymonds (approved):
LGTM

PR #1454, opened 2021-06-14
Update docker files

glattercj:
- Update docker files
- Add automatic minimega docker build workflow for Github container registry (Actions)


csymonds (approved):
LGTM

PR #1457, opened 2021-07-23
Support TCP and UDP connectivity testing from miniccc

activeshadow:
minimega, miniccc, and ron were updated to support executing tcp/udp
connectivity tests directly from miniccc agents.

ron was updated to include a new `ConnTest` command struct that
encapsulates the endpoint to test against, how long to wait, and what
UDP packet to send (if necessary).

miniccc was updated to include a handler for the new `ConnTest` command
that simply tries to dial the endpoint (in the case of TCP), and if
necessary write the UDP packet to the socket (in the case of UDP).

minimega was updated to include support for a new "cc test-conn"
command, as well as adding a "connectivity" column to the "cc commands"
table.

Documentation for minimega's "cc" command was also updated to include
details and examples of how to use the new "test-conn" command.

### DIFF
--- a/src/miniccc/commands.go
+++ b/src/miniccc/commands.go
@@ -7,11 +7,16 @@ package main
 import (
 	"bufio"
 	"bytes"
-	log "minilog"
+	"fmt"
+	"net"
+	"net/url"
 	"os/exec"
 	"path/filepath"
-	"ron"
 	"strings"
+	"time"
+
+	log "minilog"
+	"ron"
 )
 
 func processCommand(cmd *ron.Command) {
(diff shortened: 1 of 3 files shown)


activeshadow:
The reason for this new miniccc feature is to support connectivity testing without having to require certain executables to be available in target VMs and/or having to add a custom executable to VMs when the miniccc agent is already present in all the VMs. The main driver for connectivity testing is the state of health app available in phenix.

aherna (changes requested):
Having issues on connectivity test with containers, getting no responses but can other wise use miniccc @activeshadow

activeshadow:
> Having issues on connectivity test with containers, getting no responses but can other wise use miniccc @activeshadow

@aherna can you provide a little more detail about how you're testing and what you're seeing? For example, is the issue with testing connectivity to a container from a VM, or to a VM from a container, or to a container from a container? Is the target VM/container running a service on the protocol/port your testing connectivity to?

PR #1458, opened 2021-08-13
Allow user to specify which broadcast IP to use for meshage

activeshadow:
In some cases, cluster compute nodes have multiple interfaces and
networks they can communicate with each other over, and one may be
desirable over another for meshage comms. By default, minimega uses
`255.255.255.255` as the broadcast address, and when multiple interfaces
are present, the one acting as the default route is the one used.

This commit adds a new command line option, `-broadcast`, for users to
specify which broadcast IP to use for the mesh. It defaults to
`255.255.255.255`, so this commit should be backwards compatible.

### DIFF
--- a/docker/minimega.service
+++ b/docker/minimega.service
@@ -5,6 +5,7 @@ After=multi-user.target
 [Service]
 Environment="MM_BASE=/tmp/minimega"
 Environment="MM_FILEPATH=/tmp/minimega/files"
+Environment="MM_BROADCAST=255.255.255.255"
 Environment="MM_PORT=9000"
 Environment="MM_DEGREE=2"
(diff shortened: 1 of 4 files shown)


activeshadow:
@aherna anymore thoughts on this? It doesn't break normal, default behavior (i.e., it's backwards compatible) so I'm hoping you will be amenable to merging it in.

aherna:
Havent tested the with latest changes. Is this urgent or can it wait til new year?

activeshadow:
@aherna it can most definitely wait.

activeshadow:
Hi @aherna just wanted to ping you on this again.

aherna (approved):
LGTM

PR #1461, opened 2021-12-15
miniccc: fix map race conditions in ron server

activeshadow:
In commit dd04c33f I introduced a bug in the ron server code that handles serial cc connections -- failing to protect maps accessed concurrently with their associated mutexes -- that would randomly manifest itself into race conditions.


activeshadow:
@aherna from what I can tell, this seemed to fix the issues multiple people were having with larger deployments.

aherna (approved):
LGTM

PR #1462, opened 2022-02-14
Don't rely on systemd to run minimega components

activeshadow:
See https://github.com/moby/moby/issues/42275.


glattercj (approved):
LGTM

PR #1470, opened 2022-05-17
ron: version the cc message so heartbeats work across revisions

activeshadow:
This is an attempt to make miniccc somewhat future-proof. Right now, if minimega is updated then miniccc running in existing images will continually fail and reconnect because currently we only start the heartbeat on the server if the commit versions between the server and miniccc match. This is somewhat okay, except for when `cc mount` needs to be used.

By versioning the cc message, we can now accurately determine if older miniccc agents can be used with newer versions of deployed servers.


aherna (approved):
LGTM

PR #1471, opened 2022-08-04
[minimega] Prevent dnsmasq from leaking DNS results

### LINKED ISSUE #1421: Feature Request: Change default behavior of dnsmasq to prevent dns leakage
**Describe your environment**

1. minimega version latest
2. Linux distro/version Ubuntu 18.04 Server
3. Go compiler version 1.12
4. VM types Any

**Describe the bug**
The default configuration for dnsmasq is insecure and will bleed dns. I brought this up before and was told it works as intended and is not an issue. 

**To Reproduce**
Start Minimega

```
vm config net 1000
vm config cdrom /root/tinycore.iso
vm config memory 256
vm launch kvm tiny
vm start tiny
tap create 1000 ip 1.0.0.1/24
dnsmasq start 1.0.0.1 1.0.0.2 1.0.0.254
```
This should launch a VM on mega_bridge vlan 1000 and assign an ip address in 1.0.0.0/24 from the dnsmasq service running on a tap in vlan 1000.

Now hop on the VM, open a terminal and `ping google.com` you'll see an ip address resolve, but the ping fails.

This is a problem. The only way it could know an ip address is if it somehow reached out to the internet and got a response back. This virtual network may seem isolated from the internet, but is clearly not. 

Yes it is very bad practice to add this tap and not understand what it is doing. The better solution is to make sure to start dns using a vm or container in the experiment. But far too many people incorrectly assume dnsmasq is running in an isolated network because when they try and access the internet they see the internet fails. 

Run tcpdump on the server and you'll see the domain traffic going outbound.
```
tcpdump -i eth0 port 53 -vv
```
Take a look at ps aux and you'll see dnsmasq is running with the `-o` option.
```
root@ubuntu:~# ps aux  |grep dnsmasq
root       9663  0.0  0.0  59980  3992 pts/2    S+   14:51   0:00 /usr/sbin/dnsmasq -u root --pid-file=/tmp/minimega/dnsmasq_645651131/dnsmasq.pid -o -k --hostsdir=/tmp/minimega/dnsmasq_645651131/hostdir --dhcp-hostsdir=/tmp/minimega/dnsmasq_645651131/dhcpdir --dhcp-optsdir=/tmp/minimega/dnsmasq_645651131/dhcpoptsdir --except-interface lo --listen-address 1.0.0.1 --bind-interfaces --dhcp-range 1.0.0.2,1.0.0.254 --dhcp-leasefile=/tmp/minimega/dnsmasq_645651131/dnsmasq.leases --dhcp-lease-max=4294967295
```

From the man page

```
-o, --strict-order
By default, dnsmasq will send queries to any of the upstream servers it knows about and tries to favour servers that are known to be up. Setting this flag forces dnsmasq to try each query with each server strictly in the order they appear in /etc/resolv.conf
```
The default configuration of dnsmasq will resolve domains. This needs to be changed.

**Expected behavior**
We shouldn't use -o and -R should be enabled by default in dnsmasq.go. We should make a `dnsmasq configure` command to enable upstream resolving for when someone wants to use iptables forwarding.

activeshadow:
This commit updates the command line flags for dnsmasq to explicitly
point to a non-existent resolv.conf file rather than letting it use the
default /etc/resolv.conf. This prevents leakage of DNS results as
described in #1421.

This commit also adds an option to the `dns configure` command for
adding upstream DNS servers. Adding an upstream server simply adds a
nameserver entry to the (initially non-existent) resolve.conf file
dnsmasq is configured to use. Upstream servers added will immediately be
recognized by dnsmasq since it polls the file for changes.

closes #1421

### DIFF
--- a/cmd/minimega/dnsmasq.go
+++ b/cmd/minimega/dnsmasq.go
@@ -83,7 +83,8 @@ to the file.`,
 		HelpShort: "configure dhcp/dns options",
 		HelpLong: `
 Configuration options for running dnsmasq instances. Define a static IP
-allocation, specify a hostname->IP mapping for DNS, or set DHCP options.
+allocation, specify a hostname->IP mapping for DNS, configure upstream DNS
+servers (useful when forwarding/NAT is enabled), or set DHCP options.
 
 To list all existing static IP allocations on the first running dnsmasq
 server, do the following:
@@ -102,6 +103,14 @@ To see DNS entries:
 To add a DNS entry:
 
 	dnsmasq configure 0 dns 172.17.0.50 example.com
+
+To see upstream DNS servers:
+
+	dnsmasq configure 0 upstream
+
+To add an upstream DNS server:
+
+	dnsmasq configure 0 upstream server 1.1.1.1
 
 To see a list of all DHCP options:
 
--- a/cmd/minimega/dnsmasq.go
+++ b/cmd/minimega/dnsmasq.go
@@ -21,16 +21,17 @@ import (
 )
 
 type dnsmasqServer struct {
-	Addr        string
-	MinRange    string
-	MaxRange    string
-	Path        string
-	Hostdir     string
-	DHCPdir     string
-	DHCPoptsdir string
-	DHCPhosts   map[string]string // map MAC to IP address
-	Hostnames   map[string]string // map IP to hostname
-	DHCPopts    []string          // DHCP options
+	Addr            string
+	MinRange        string
+	MaxRange        string
+	Path            string
+	Hostdir         string
+	DHCPdir         string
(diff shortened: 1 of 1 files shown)


jacdavi:
I think it would be beneficial to add some info about setting the upstreams in the help docs. Otherwise, everything looks good to me.

activeshadow:
> I think it would be beneficial to add some info about setting the upstreams in the help docs. Otherwise, everything looks good to me.

@jacdavi let me know if my latest commit is sufficient.

jacdavi:
Yeah I think that's perfect. Thanks!

aherna (approved):
LGTM

PR #1472, opened 2022-08-09
[minimega] Get files to be sent to miniccc via mesh before sending

activeshadow:
When a file is sent to a VM using cc, it's expected that the file is
already on the node running the VM. This commit adds support for
automatically getting the file to be sent to a VM from other nodes in
the mesh, similar to how it's done for VM disk images when they are
launched.

### DIFF
--- a/cmd/minimega/cc_cli.go
+++ b/cmd/minimega/cc_cli.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


jacdavi:
Looks good to me @aherna

aherna (approved):
LGTM

jacdavi:
Hey @activeshadow , I just noticed that doing a `cc send` seems to fail if there's a wildcard (e.g., `cc send *.txt`). I'm wondering if it's related to this PR since it used to work, but I'm not sure if `iomHelper` handles wildcards. Thoughts?

activeshadow:
@jacdavi I specifically remember testing wildcard sends for this PR. However, I'm wondering if maybe #1477 broke it? At any rate, maybe open up a new issue for it and I can take a look when I have time in the next few days.

PR #1473, opened 2022-08-10
[minimega] Add flag for sending logs to a node via the mesh

activeshadow:
The new flag is `-lognode` and takes the name of a node on the mesh.
When minimega is deployed to other nodes using `deploy launch` the flag
will automatically be included using the hostname the deployment is
being run from.


aherna (approved):
LGTM

PR #1475, opened 2022-08-23
Prevent Locking when Downloading File in miniweb

jacdavi:
Previously, downloading a file on miniweb would lock minimega and the UI for the duration of the download. If the download was cancelled, things would remain locked for an extended period of time.

This PR makes miniweb create a new connection to minimega for each file download. This allows miniweb to still send other commands through its normal channel while handling a file. This also fixes the issue of a cancelled download as the new connection will simply be closed. You can also now download multiple files at the same time.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -5,6 +5,7 @@ Devin Cook <devcook@sandia.gov>
 Miles Crabill <mcrabil@sandia.gov>
 Jonathan Crussell <jcrusse@sandia.gov>
 Jerry Cruz <gcruz@sandia.gov>
+Jacob Davis <jacdavi@sandia.gov>
 Jeremy Erickson <jericks@sandia.gov>
 John Floren <jfflore@sandia.gov>
 David Fritz <djfritz@sandia.gov>
--- a/cmd/minimega/cc_cli.go
+++ b/cmd/minimega/cc_cli.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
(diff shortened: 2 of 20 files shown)


activeshadow:
@aherna @jacdavi just FYI, there will be a conflict between this PR and my PR #1474, since my PR removes the use of `cmdLock` completely in lieu of a channel to serialize commands.

jacdavi:
@activeshadow Yeah I was aware that I'll probably need to rewrite the changes, but wanted to get the PR posted.

For your PR, I was curious how it works with the [read api](https://github.com/sandia-minimega/minimega/pull/1474/files#diff-646d6ca3e5937fc071b97ef4de8093ef4c0d5530ddfad898d7ac1cc70a1d3d88L181) since I used the same pattern here. I'm not super familiar with Go yet, so it wasn't clear to me how you avoided blocking on that call

activeshadow:
@jacdavi by using a channel instead of a mutex, commands added to the channel get processed in a FIFO fashion. This means different commands from different goroutines (read API, miniweb requests, etc) can be intertwined in the cmd channel without blocking or having to worry about locking. It's the whole "share memory by communicating" philosophy.

aherna:
@jacdavi Feel free to resolve conflict

jacdavi:
@aherna ok, should be all set. Thanks to @activeshadow only the miniweb side required changes now.

aherna:
@jacdavi this all set so merge in?

jacdavi:
yep!

aherna (approved):
LGTM

PR #1476, opened 2022-08-24
[minimega] Periodically publish status of file transfers

activeshadow:
A new "status update" construct was added to provide long running
commands the ability to periodically send status updates about the long
running command to one or more nodes in the mesh.

A "status update" API command was added to allow users to adjust how
often, if at all, status updates are sent.

A developer article was added to help developers understand how the
construct was added and how it can be used.

While testing status updates, it was discovered that a lot of error logs
were being generated for things that weren't actually errors, so that
was fixed as well.


aherna (approved):
LGTM

PR #1479, opened 2022-09-21
[miniccc] Track and record exit code for exec'ed commands

activeshadow:


jacdavi (approved):

jacdavi:
@activeshadow lmk if you have any issues with the change I made in [2e6bae5](https://github.com/sandia-minimega/minimega/pull/1479/commits/2e6bae5827efdf2d98c1754f5e4046ecaac254cc). Based on the description I expected it to work with the vm name, but that wasn't originally the case

activeshadow:
> @activeshadow lmk if you have any issues with the change I made in [2e6bae5](https://github.com/sandia-minimega/minimega/pull/1479/commits/2e6bae5827efdf2d98c1754f5e4046ecaac254cc). Based on the description I expected it to work with the vm name, but that wasn't originally the case

@jacdavi yeah it's been too long. 😂 So what's the difference between "vm name" and "hostname" now?

activeshadow:
Based on how the "FindVMs" function works, perhaps we should use "<vm id, name, or uuid>" for the command args?

jacdavi:
So what you had was matching by UUID or [hostname](https://github.com/sandia-minimega/minimega/pull/1479/files#diff-f718901cfa0fbf2e49631335be596b54bfaebd3fafeb17bf6173659b5ab59fa8R378) where the hostname is the VM's actual hostname as reported by miniccc. Whereas `FindVMs` will match by the vm name used when creating the VM (e.g., `vm launch kvm <name>`).

activeshadow:
@jacdavi ah yes. I remember now. 😂 So in reality, the VM ID, name, or UUID, or the hostname as reported by miniccc should work... should we worry about showing the VM ID should work too, or what the difference is between VM name and hostname?

jacdavi:
I guess I mostly copied `vm mount` which mentions it works by name or UUID, though ID would work too. So it's fine to leave that off. 

I'm not sure how to further clarify the vm name vs hostname. There's also a potential issue there where multiple VMs can have the same hostname (which would be poor design, but not technically wrong).

activeshadow:
LGTM. Let me know if you want me to squash and rebase.

PR #1483, opened 2022-11-22
[igor] Support Cumulus switches

activeshadow:


jacdavi:
Closing since igor development has moved to https://github.com/sandia-minimega/igor2 and we won't be making further changes here.

PR #1484, opened 2022-11-22
[igorweb] Add node info handler that includes user ssh keys

activeshadow:
This new endpoint in igorweb allows for reserved nodes to make a call to igorweb to get reservation details and ssh keys for users that have access to the reservation.


jacdavi:
Closing since igor development has moved to https://github.com/sandia-minimega/igor2 and we won't be making further changes here.

PR #1486, opened 2022-11-28
[bridge] Support for bonded interfaces

activeshadow:
* [x] Add `bridge bond` command and corresponding function to internal bridge.
* [x] Add `clear bridge bond` command and corresponding function to internal bridge.
* [x] Update `bridge.DestroyTap` and `bridge.RemoveTap` to remove bond ports of no members are left.
* [x] Update `bridge` command to show bond details for bridge.
* [x] Update `vm info` command to show bond details for VM.


activeshadow:
@aherna this is ready for review and testing.

aherna:
awesome

> @aherna this is ready for review and testing.
> 
> Also, be sure to let me squash all the commits before merging the PR. I'll do that after you're done with your review and testing.

Ill test this out over the next few days

aherna:
@activeshadow 
See below scenario
made a bond for VM r1 for ports 1,2 then cleared a bond
vm info shows
host    |vm  | vlans                                  | taps
hosta  | r1   | [a (101), b (102), c (103)] | [mega_tap9, mega_tap10, mega_tap11]

upon inspection in ovs-vsctl show looks like the taps got put back but the vlan tags are incorrect, 
both mega_tap10 and 11 show tag of 102 when they should show 102 and 103 respectively

internal/bridge/bond.go
```diff
@@ -0,0 +1,114 @@
...
+	// ovs-vsctl add-bond <bridge name> <bond name> <list of interfaces>
+	args := []string{"add-bond", b.Name, name}
+
+	args = append(args, interfaces...)
+	args = append(args, "--", "set", "port", name)
+	args = append(args, fmt.Sprintf("tag=%v", vlan))
+
+	// https://www.man7.org/linux/man-pages/man5/ovs-vswitchd.conf.db.5.html#Port_TABLE
+	switch mode {
+	case "active-backup", "balance-slb":
```
aherna, line 32: for the different bond modes you can ahve LACP active,passive,or off in our case we want to specifiy lacp, maybe we had a lacp field to the bridge cli?

activeshadow: @aherna okay, so I must have misunderstood some of the documentation. I thought LACP was only applicable for bond mode `balance-tcp`. Sounds like LACP is applicable for all bond modes that OVS supports.

I can add the extra bridge API command to activate LACP for a bond no problem. Do you want to automatically enable LACP Active/Backup Fallback Mode when LACP is active, or do you want to make that configurable as well?

activeshadow:
> @activeshadow See below scenario made a bond for VM r1 for ports 1,2 then cleared a bond vm info shows host |vm | vlans | taps hosta | r1 | [a (101), b (102), c (103)] | [mega_tap9, mega_tap10, mega_tap11]
> 
> upon inspection in ovs-vsctl show looks like the taps got put back but the vlan tags are incorrect, both mega_tap10 and 11 show tag of 102 when they should show 102 and 103 respectively

@aherna okay I made a bad assumption here. I think I can fix it pretty easily.

internal/bridge/bond.go
artherna: I would just add the optional lacp parameter to the current setup and default to active lacp for all bond types

artherna:
@activeshadow i think theres a miscommunication somewhere.

there are three bond modes and there are three lacp modes

Lacp can be active, passive, or off and special lacp option for fallback has to be enabled with one of the bond modes.

basically what you had before was fine just wanted an addition of an lacp field 

ie
Bridge bond <bondname> <bridge> <active-backup,balance-slb,…> [active,passive,off]

bridge bond foobond mega_bridge balance-slb active
bridge bond foobond mega_bridge balance-slb passive
bridge bond foobond mega_bridge balance-slb off

the user can ommit lacp field and just leave lacp active by default.

if you look at the link posted
https://www.man7.org/linux/man-pages/man5/ovs-vswitchd.conf.db.5.html#Port_TABLE

activeshadow:
@aherna I'm causing problems here, sorry! I'll get the LACP mode added as an explicit option to the API.

Note that for balance-tcp, LACP must be enabled (either passive or active) so I'll error out if someone tries to turn it off. 

Also, the LACP fallback option is applicable to all modes so I'll leave it enabled by default but still provide a way to disable it.

activeshadow:
@aherna okay give that last commit a go and let me know what you think.

aherna:
@activeshadow alright i did a more through test and it looks great*
* = works on single node only

we need to fix this so that it works in mesh. its not creating the bonds correctly on remote hosts ie when it tries to create the bonds it looks for the vm locally

activeshadow:
@aherna okay this is working now in a mesh.

activeshadow:
@aherna latest commits support QinQ for bonded interfaces. I've tested on a mesh. Can you (or do you want to) test one more time before I squash?

artherna:
Ill do one more test within 30 min

aherna:
@activeshadow looks good, one thing i saw is qinq field in vm info is blank even though bond is set to have qinq. will need to adjust so that qinq shows up as true for that bond. I think it's because we only set true if it's done at interface level.
host1 | r2   | []   | [mega_bond5 (mega_tap135 mega_tap136), mega_bond4 (mega_tap133 mega_tap134)]

cmd/minimega/vm.go
```diff
@@ -314,6 +322,105 @@ func (vm *BaseVM) GetNetworks() []NetConfig {
...
+		return err
+	}
+
+	for _, c := range ifaces {
+		if c.QinQ {
+			// TODO: Should we ignore errors here? Or should we return immediately if
+			// an error occurs? Or should we gather up multiple errors, go ahead and
+			// delete the bond from "vm.Bonds", and then return the multiple errors as
+			// a single error?
+			br.SetTapQinQ(c.Tap, c.VLAN)
```
aherna, line 416: we should try to cleanup if we can and return single error, and/or show as warning with the collection being an error?

activeshadow: Okay @aherna can do. I'll work on that now.

activeshadow:
> @activeshadow looks good, one thing i saw is qinq field in vm info is blank even though bond is set to have qinq. will need to adjust so that qinq shows up as true for that bond. I think it's because we only set true if it's done at interface level. In teh example below bond4 is qinq and bond5 has no qinq host1 | r2 | [] | [mega_bond5 (mega_tap135 mega_tap136), mega_bond4 (mega_tap133 mega_tap134)]

Whoops okay @aherna I'll fix this.

activeshadow:
@aherna okay both are fixed. Do you want to test again or do you want me to squash so you can merge?

artherna:
Ill do a last pass

aherna:
@activeshadow LGTM squash and ill merge it in!

activeshadow:
@aherna commits have been squashed. :+1:

aherna (approved):
Tested, LGTM

PR #1490, opened 2023-01-25
[minimega] Retry (with backoff) finding disk partitions for injects

activeshadow:
Lately we've been seeing a lot of `desired partition <n> not found` errors when injecting files into VM disk images. Adding this retry with backoff has proven to help reduce how often we see this issue (basically we don't see it anymore at all).

### DIFF
--- a/cmd/minimega/disk.go
+++ b/cmd/minimega/disk.go
@@ -227,11 +227,21 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
 		path = nbdPath + "p" + partition
 
 		// check desired partition exists
-		_, err = os.Stat(path)
-		if err != nil {
-			return fmt.Errorf("[image %s] desired partition %s not found", dst, partition)
-		} else {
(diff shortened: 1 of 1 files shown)


jacdavi:
This seems to work as expected, but I'm only testing without the error. Is there a way to get the `desired partition <n> not found` errors?

activeshadow:
@jacdavi you can force the error by specifying a non-existent partition when injecting files, but that won't really test the retry. We were seeing this error a lot, even when the partition existed, when we were deploying a very large number of VMs. It was as though the code was connecting and mounting a large number of disk images very fast, and disconnecting them wasn't happening fast enough before the next one was attempted to be connected.

I get why you're trying to test, but I can confirm this code still works... we've had no trouble with it since we've started using it back in late January of this year.

jacdavi:
@activeshadow thanks for the additional context. Code itself looks good to me, and since you've had multiple people using it without issue, I'm good to merge.

jacdavi (approved):

PR #1492, opened 2023-02-03
[minimega] Upgrade version of google/gopacket module

activeshadow:
Started hitting the following bug in some larger deployments.

https://github.com/google/gopacket/issues/763

Upgrading to v1.1.19 pulls in the fix for the above issue.


jacdavi:
Covered by #1507

PR #1500, opened 2023-04-03
Add config option for switching from xHCI to EHCI for USB controller

jacdavi:
This PR adds a configuration option to switch between xHCI and EHCI for the USB controller. Only one will be enabled per VM and by default xHCI is used.

To disable xHCI and use EHCI instead (as before), add `vm config usb-use-xhci false` to the VM's config.

This should fix two issues:
1. Issues observed for Windows VMs in phenix when both EHCI and xHCI are present
2. Backwards compatibility for certain use cases that cannot add support for xHCI (e.g., pre-2010 VMs)

### DIFF
--- a/cmd/minimega/kvm.go
+++ b/cmd/minimega/kvm.go
@@ -179,6 +179,12 @@ type KVMConfig struct {
 	// Note: this configuration only applies to KVM-based VMs.
 	Disks DiskConfigs
 
+	// If true will use xHCI USB controller. Otherwise will use EHCI.
+	// EHCI does not support USB 3.0, but may be used for backwards compatability.
+	//
+	// Default: true
+	UsbUseXHCI bool
+
 	// Add additional arguments to be passed to the QEMU instance. For example:
 	//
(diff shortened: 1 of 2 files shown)


jacdavi:
@causand22 could you test this out? I haven't done any testing myself yet, but plan to tomorrow

causand22:
@jacdavi 

Tested the following:
- Launched a VM with  `vm config usb-use-xhci true`
  - hotplug USB 3.0 works: hotplug command works, drive can be mounted within vm.
- Launched a VM with `vm config usb-use-xhci false`
  - hotplug usb 3.0 throws error `Error (ccc5.40g): invalid version: `3.0`. VM configured for EHCI instead of XHCI`
  - hotplug usb 2.0 works: hotplug command works
- Launched VM without specifying `usb-use-xhci`
  - Same behavior as specifying `true`

Looks good to me.

jacdavi:
Closing in favor of #1501 since I can't approve this as the creator...

PR #1509, opened 2023-06-29
[minitunnel] track tunneled ports for listing and closing

activeshadow:
Track and display port forward tunnels created using miniccc, and allow for existing tunnels to be torn down.

### DIFF
--- a/cmd/minimega/cc_cli.go
+++ b/cmd/minimega/cc_cli.go
@@ -97,6 +97,11 @@ without arguments displays the existing mounts. Users can use "clear cc mount"
 to unmount the filesystem of one or all VMs. This should be done before killing
 or stopping the VM ("clear namespace <name>" will handle this automatically).
 
+"cc tunnel" allows users to tunnel TCP connections to a local port through a VM
+to a remote port. The local port will be created on the minimega cluster host
+that the tunneling VM is running on. The remote port can be on the same VM or on
+a different VM the tunneling VM has network access to.
+
 "cc test-conn" allows users to test network connectivity from a guest to the
 given IP or domain name and port. The wait timeout should be specified as a Go
 duration string (e.g. 5s, 1m). If "udp" is used, a "base64 udp packet" that will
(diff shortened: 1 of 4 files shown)


jacdavi:
Yep I can take a look next week. Trying to get version 2.8 released atm, so this will be the first change for 2.9 :)

activeshadow:
Ah shoot... that means this one and the one I created earlier today missed the cutoff. 😞

cmd/minimega/cc_cli.go
jacdavi, line 1: Could you add a small blurb in the long help about `cc tunnel`?

cmd/minimega/cc_cli.go
```diff
@@ -236,6 +238,85 @@ func cliCC(ns *Namespace, c *minicli.Command, resp *minicli.Response) error {
...
+
+	if c.BoolArgs["list"] {
+		// map VM name --> VM UUID
+		vms := make(map[string]string)
+
+		if v == Wildcard {
+			clients := ns.ccServer.GetClients()
+
+			for _, client := range clients {
+				vms[client.Hostname] = client.UUID
```
jacdavi, line 270: Small issue I noticed here. I had two VMs booted that had the same hostname, so they clobbered each other, and only 1 showed. It's also slightly inconsistent that `all` outputs hostnames, while you need the actual vm name or UUID when specifying a specific vm.

Maybe we do an extra lookup to convert from UUID -> name then use that as the key?

jacdavi (commented):
One final thing. On closing a tunnel, I consistently get error outputs from the closed connection

```
2023/08/24 19:31:30 ERROR minitunnel.go:219: accept tcp [::]:9023: use of closed network connection
2023/08/24 19:31:30 ERROR minitunnel.go:303: read tcp [::1]:9023->[::1]:39264: use of closed network connection
```
(only got the 2nd line once)

Is it possible to suppress those errors when the user explicitly closed the tunnel?

---

Otherwise this is really cool! I hadn't used `cc tunnel` before, and I think if I had, I would've expected these features to already be there.

cmd/minimega/cc_cli.go
activeshadow: Please take a look at my latest commits that attempt to address this and your comment above.

cmd/minimega/cc_cli.go
activeshadow: My latest commits should address this as well.

jacdavi (approved):

jacdavi:
Looks good! Thanks for the changes!

PR #1515, opened 2023-08-10
[minitest] fix some tests

gelaski:
Updated several test `.want` files as well as some inconsistent behavior in minimega.

Test that are still failing:
`vm_add_nic`
tests in `tests/distributed`

### DIFF
--- a/tests/cc_io
+++ b/tests/cc_io
@@ -3,7 +3,7 @@ background dd if=/dev/zero of=/tmp/minimega/files/cc_io.bigfile bs=1M count=512
 
 vm config filesystem $images/minicccfs
 vm config uuid 9e73ce1b-b93d-4594-8577-80d773b1fbb9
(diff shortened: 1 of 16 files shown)


tests/cc_io
```diff
@@ -3,7 +3,7 @@ background dd if=/dev/zero of=/tmp/minimega/files/cc_io.bigfile bs=1M count=512
 
 vm config filesystem $images/minicccfs
 vm config uuid 9e73ce1b-b93d-4594-8577-80d773b1fbb9
-vm config memory 512
+vm config memory 1024
```
jacdavi, line 6: It looks like the intent of the test, based on line 1 and prior commits is for the file to exceed the vm's memory size. So not sure this is the right solution

jacdavi:
Looks good. Will merge into `minitest-fixes`

PR #1516, opened 2023-08-11
[minimega] fix bridge variable so deleting netflow capture works

activeshadow:
When deleting a netflow capture, the bridge name was getting assigned to the `<name>` variable but the code that deletes the capture was trying to get the name of the bridge from the `<bridge>` variable.


jacdavi (approved):

activeshadow:
@jacdavi can you merge this in since it's been approved?

PR #1520, opened 2023-10-02
Add copy/paste support

jacdavi:
Adds two levels of copy/paste support:

1. Primitive Paste - enabled by default for all VMs. On a `ClientCutText` VNC request, types out the contents of the paste buffer
2. Bidirectional Copy/Paste - can be set by the `bidirectionalCopyPaste` vm config. When enabled, copy/paste does not type directly, but interacts with the host and VM clipboards directly. Note that this setting requires QEMU 6.1+ compiled with the vdagent and for `spice-vdagent` to be installed on the VM.

Additionally adds a new `vnc type` command which will type the provided ascii string into the specified vm. All shifted characters on a US keyboard are handled correctly. e.g., `vnc type my-vm "this is a long string I want in the VM!!"`. Under the hood, this is what the primitive paste is utilizing.


aherna (changes requested):
LGTM, Is there anyway to make this backwards compatible? IE detect version of QEMU disable copy paste for older qemu

jacdavi:
> LGTM, Is there anyway to make this backwards compatible? IE detect version of QEMU disable copy paste for older qemu

Does it work to check at vm start and just disable the bidirectional copy/paste with a warning if an old QEMU is used? The way it is now, it's backwards compatible by default, but setting `vm config bidirectionCopyPaste true` requires QEMU 6.1+

PR #1524, opened 2023-12-14
Bump version to 2.9; add global version file

jacdavi:
* use version file for deb and python packages
* print version number in minimega (in addition to commit/date)
* Build python dist if `python3` is present as well (used to only check if `python` was available)


jacdavi:
Marking as draft since we want to merge this one last for 2.9

aherna (approved):
LGTM

PR #1541, opened 2024-08-14
Add command to print arguments in use

jacdavi:
Adds a command to print the CLI arguments minimega is using. 

This is being added so other programs (e.g., phenix) have any easy way to identify paths, ports, etc. Specifically this will replace phenix using `ps` to determine the `filepath`: https://github.com/sandialabs/sceptre-phenix/blob/main/src/go/util/file.go#L32


aherna (approved):
LGTM

PR #1545, opened 2024-10-14
Add new disk commands

jacdavi:
* Adds commands for manipulating disks (wrappers for `qemu-img`)
  * `disk <rebase,> <image> [backing file]`
  * `disk <set-backing,> <image> [backing file]` (rebase with unsafe flag)
  *  `disk resize <image> <size>`
  * `disk commit <image>`
* Updates `disk info` with `recursive` option which prints entire backing chain
* Updates `disk info` with new column for "in use"
  * This is tracked by checking if qemu has a lock on the disk file. This will mark backing images of booted images as in use
* Change `snapshot` to use a backing file path relative to the base image rather than an absolute path
  * Change `disk info` output for backingfile to show relative path if backing file is relative
* Minor code restructure for readability
* Parse `qemu-img info` as json rather than regex on a plain string
* Format disk size as human readable IEC format


activeshadow:
@glattercj we should look to see if any of these updates will allow us to pull some code out of phenix (that was perhaps doing some of the same stuff as these updates are now doing lower in the stack).

aherna (approved):
Lgtm

jacdavi:
> @jacdavi I'm curious of the reasoning behind changing the `snapshot` command to use relative backing file paths? I've come across two different use cases now where the use of relative paths has broken an existing workflow: creation of new backing images and restoration of namespace snapshots.
> 
> See [sandialabs/sceptre-phenix#246](https://github.com/sandialabs/sceptre-phenix/issues/246) for details on the backing image issue. I don't think using the `disk rebase` command added via this PR will help in this case since we move the copy of the minimega snapshot we make before rebasing, but I haven't tested it to confirm.
> 
> The issue I'm seeing with namespace snapshot restoration happens when a namespace is deleted after the snapshot is created. The snapshot disks for the VMs in the namespace have a backing file path that is a relative path from the minimega snapshot disk in `/tmp/minimega/{id}` but if the namespace is deleted those paths no longer exist.
> 
> It seems like the use of absolute paths for backing files would prevent both of these from happening...

This change was motivated by user-managed images. We had a need/desire to start using snapshots to manage our topology images as we iterated them. Absolute paths were troublesome when moving to environments with different pathing (e.g., between our local and docker deployments) and the errors were unintuitive and not quick to fix.

I agree though that in automated cases using paths like above, it makes more sense to use absolute paths. Maybe the best solution is to add a flag to the snapshot command for the path to be absolute?

sdelliot:
> > This change was motivated by user-managed images. We had a need/desire to start using snapshots to manage our topology images as we iterated them. Absolute paths were troublesome when moving to environments with different pathing (e.g., between our local and docker deployments) and the errors were unintuitive and not quick to fix.
> > I agree though that in automated cases using paths like above, it makes more sense to use absolute paths. Maybe the best solution is to add a flag to the snapshot command for the path to be absolute?
> 
> @jacdavi yes, I like that idea! Even better, we could consider also adding a startup flag that would set the default if a flag to the snapshot command is not provided.
> 
> /cc @glattercj

Adding on to this that I am just running into this issue as well...We are running into it when using the `ns save` command and then restarting minimega (e.g., simulated reboot). Then when starting everything back up by running the saved `launch.mm` script, a lot of the VMs don't boot correct due to this relative path issue.

jacdavi:
@sdelliot @activeshadow If someone is able to create a PR to address this, I'd be happy to merge. I don't have any availability to develop it myself :(

PR #1546, opened 2024-11-19
Add new background commands

causand22:
Adding cli commands to view the status of `background` shell commands. When a call to `background [cmd]` is executed, mm stores the associated command alongside a sequential ID and some information (time start, time end, stdout, stderr, running). When the process exits, the stdout and stderr outputs are saved and the process is marked as complete. Users can then grab the stdout or stderr of a completed process. There is a limit of 50 background processes stored in the status table. When the 51st is created, the oldest (sorted by time of the process ending) will be removed from the table and stdout / stderr can no longer be accessed using `background-output`. All outputs of background processes are also logged to the logger on level info. 

Commands changed:
- `background`
  - Now returns an ID for the background process. Creates an entry in the background process table.

Commands added:
- `background-status [id]`
  - Get the status of all the background processes in a tabular format. 
  - If `id` is provided, only return the status of that entry (if exists)
- `clear background-status`
  - Deletes all background processes from the table that are done running (freeing up memory for saved stdout/stderr)
- `background-output <id>`
  - Get raw stdout from a specific background process
- `background-error <id>`
  - Get raw stderr from a specific background process

### DIFF
--- a/cmd/minimega/shell.go
+++ b/cmd/minimega/shell.go
@@ -148,6 +148,7 @@ func (bp BackgroundProcess) ToTabular() []string {
 	}
 	return []string{
 		strconv.FormatInt(int64(bp.ID), 10),
+		strconv.FormatInt(int64(bp.Command.Process.Pid), 10),
 		strconv.FormatBool(bp.Running),
 		errorsString,
 		bp.TimeStart.Format("Jan 02 15:04:05 MST"),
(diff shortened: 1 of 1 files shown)


jacdavi (approved):
Tested out and read through the code and everything looks good. 

I am thinking of a scenario though where a background command gets stuck and a user wants to manually clear it. We could add some way to kill it, but probably sufficient and easier to just add the `PID` to the table (`strconv.FormatInt(int64(bp.Command.Process.Pid), 10)`)

PR #1547, opened 2024-11-19
Docker and Python build enhancements

GhostofGoes:
Misc. Docker and Python improvements
- Add standardized metadata to the Docker image, including the version, commit hash, and build timestamp. This makes it easier to determine what the container was actually built from.
- Cleanup docker/README. Remove old unsupported `docker-compose` and replace with modern `docker compose`, add some missing variables, add some aliases, and some formatting tweaks.
- Add and cleanup Python package metadata in `setup.py`, notably README formatting, and `python_version` (which helps pip and dependency resolvers like PDM and Poetry).


GhostofGoes:
This should be ready to merge, pending review. The GitHub Actions pipelines pass in my fork.

docker/Dockerfile
```diff
@@ -9,14 +14,34 @@ COPY . .
...
+  org.opencontainers.image.documentation="https://www.sandia.gov/minimega/" \
+  org.opencontainers.image.source="https://github.com/sandia-minimega/minimega" \
+  org.opencontainers.image.version="${MM_VERSION}" \
+  org.opencontainers.image.revision="${MM_COMMIT}" \
+  org.opencontainers.image.vendor="Sandia National Laboratories" \
+  org.opencontainers.image.licenses="GPL-3.0-only" \
+  org.opencontainers.image.title="minimega" \
+  org.opencontainers.image.description="minimega is a tool for launching and managing virtual machines"
+
+ENV DEBIAN_FRONTEND="noninteractive"
```
jacdavi, line 36: This line is a duplicate, so should be good to remove

docker/README.md
```diff
@@ -46,47 +50,44 @@ docker run -d \
...
 # Extras
 
-### Convenience aliases
+## Convenience aliases
 
 ```bash
 cat <<EOF >> ~/.bash_aliases
-alias minimega='docker exec -it minimega minimega '
+alias mm='docker exec -it minimega minimega -e'
+alias mminfo='mm .columns name,state,ip,snapshot,cc_active vm info'
```
jacdavi, line 76: Maybe want `vm info summary` here, but this is fine too

jacdavi:
Forgot to write yesterday, but besides the few very minor changes I think this looks great! I think docker and python tend to get a bit neglected in the repo, so I appreciate the enhancements.

GhostofGoes:
Thanks for the review! I'll address the changes once I'm back in the office next week.

docker/Dockerfile
GhostofGoes: Line 7 runs in context of gobuilder image build (line 5, `FROM golang:1.18.2 AS gobuilder`), while line 36 is in context of the minimega image build (which starts at line 18).

docker/README.md
GhostofGoes: `vm info summary` doesn't show IPs, but i'll add a new alias

.github/workflows/docker.yml
```diff
@@ -9,24 +9,31 @@ jobs:
       contents: read
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: Get short SHA
         id: short
```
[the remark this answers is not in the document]
GhostofGoes: renamed the step and id

GhostofGoes:
@jacdavi Fixes for review comments are pushed and it's ready for final review

docker/Dockerfile
jacdavi: Ah gotcha. Wasn't familiar with multi-stage builds, but that makes sense

docker/README.md
jacdavi: That's fine. Just interesting to see your preferred columns vs what's in `vm info summary`. Made me wonder if summary should be updated.

jacdavi (approved):

docker/README.md
glattercj: @jacdavi I added `mminfo` as an alias to our internal scripts many years ago before `vm info summary` existed, and those were the fields we were most interested in. We just never switched over to summary and mminfo stuck around this whole time :) IMHO `id,name,state,ip,vlan,cc_active` would be a great summary nowadays, but it's pretty subjective and pry not worth code changes since it's easy to add aliases.

PR #1548, opened 2024-11-22
feat: allow disk inject to delete files

glattercj:
Adds the ability to delete files while doing disk inject via a keyword as the src. E.g.:

`disk inject window7_miniccc.qc2 files "-DELETE-":"Program Files/miniccc.exe"`

### DIFF
--- a/cmd/minimega/disk.go
+++ b/cmd/minimega/disk.go
@@ -78,14 +78,21 @@ should be quoted. For example:
 
 	disk inject foo.qcow2 options "-t fat -o offset=100" files foo:bar
 
+To delete files or directories from an image, the src should be specified
+as "-DELETE-", with dst being the file or directory to delete from the image.
+For example:
+
+	disk inject window7_miniccc.qc2 files "-DELETE-":"Program Files/miniccc.exe"
+	disk inject window7_miniccc.qc2 files "-DELETE-":"Users/Administrator/Documents/TestDir"
+
 Disk image paths are always relative to the 'files' directory. Users may also
 use absolute paths if desired. The backing images for snapshots should always
 be in the files directory.`,
 		Patterns: []string{
 			"disk <create,> <qcow2,raw> <image name> <size>",
 			"disk <snapshot,> <image> [dst image]",
-			"disk <inject,> <image> files <files like /path/to/src:/path/to/dst>...",
-			"disk <inject,> <image> options <options> files <files like /path/to/src:/path/to/dst>...",
+			"disk <inject,> <image> files <files like /path/to/src|-DELETE-:/path/to/dst>...",
+			"disk <inject,> <image> options <options> files <files like /path/to/src:/path/to/dst>",
 			"disk <info,> <image>",
 		},
 		Call: wrapSimpleCLI(cliDisk),
@@ -282,14 +289,22 @@ func diskInject(dst, partition string, pairs map[string]string, options []string
 		}
 	}()
 
-	// copy files/folders into mntDir
+	// copy files/folders into mntDir. If src matches the keyword
+	// '-DELETE-', delete the dst from mntDir.
 	for target, source := range pairs {
-		dir := filepath.Dir(filepath.Join(mntDir, target))
-		os.MkdirAll(dir, 0775)
+		if source == "-DELETE-" {
+			err := os.RemoveAll(filepath.Join(mntDir, target))
+			if err != nil {
+				return fmt.Errorf("[image %s] error deleting '%s': %v", dst, target, err)
+			}
+		} else {
+			dir := filepath.Dir(filepath.Join(mntDir, target))
+			os.MkdirAll(dir, 0775)
 
-		out, err := processWrapper("cp", "-fr", source, filepath.Join(mntDir, target))
-		if err != nil {
-			return fmt.Errorf("[image %s] %v: %v", dst, out, err)
+			out, err := processWrapper("cp", "-fr", source, filepath.Join(mntDir, target))
+			if err != nil {
+				return fmt.Errorf("[image %s] %v: %v", dst, out, err)
+			}
 		}
 	}
 
--- a/cmd/minimega/disk.go
+++ b/cmd/minimega/disk.go
@@ -78,12 +78,13 @@ should be quoted. For example:
 
 	disk inject foo.qcow2 options "-t fat -o offset=100" files foo:bar
 
-To delete files or directories from an image, the src should be specified
-as "-DELETE-", with dst being the file or directory to delete from the image.
-For example:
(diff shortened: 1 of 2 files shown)


jacdavi (approved):
Tested and looks good. Two UX considerations though:

* Would it make sense to have the syntax be `disk inject <image> delete files <files like /path/to/dst>`? I realize it would require a bit more refactoring, but on the surface it seems more intuitive to me
* Would it be useful to log a warning if a file the user tries to delete doesn't exist? I fat-fingered a path in testing, and wouldn't have noticed if I wasn't manually verifying the image after. `os.RemoveAll` just returns `nil` in that case, so we'd have to separately check if the file exists.

glattercj:
@jacdavi 

- Updated to use your suggested format. E.g.:

    ```bash
    disk inject window7_miniccc.qc2 delete files "Program Files/miniccc.exe"
    ```

- Added warning if path doesn't exist for deletion

jacdavi:
Thanks I think those are both really good changes! It does seem like there's an issue parsing when deleting a single file. Based on the docstring I think this should work?

```
minimega[minimega]$ disk inject /scratch/images/test.qcow2 delete files /etc/setup.sh
Error (ccc19.40g): malformed command; expected src:dst pairs
```

glattercj:
> Thanks I think those are both really good changes! It does seem like there's an issue parsing when deleting a single file. Based on the docstring I think this should work?
> 
> ```
> minimega[minimega]$ disk inject /scratch/images/test.qcow2 delete files /etc/setup.sh
> Error (ccc19.40g): malformed command; expected src:dst pairs
> ```

@jacdavi Good catch, I missed the single file case 🤦 . Updated and tested.

cmd/minimega/disk.go
```diff
@@ -343,6 +355,37 @@ func parseInjectPairs(files []string) (map[string]string, error) {
...
+	case []string:
+		if !(sliceContainsCharacters(v, ",:")) {
+			paths = []string{v[0]}
+		} else if sliceContainsString(v, ",") {
+			paths = strings.Split(v[0], ",")
+		} else {
+			pairs, err = parseInjectPairs(v)
+		}
+	case string:
+		if strings.Contains(v, ",") {
```
jacdavi, line 377: Can the single case also be added here. Otherwise it doesn' t work with options specified

```
minimega[minimega]$ disk inject /scratch/images/test.qcow2 options "-o offset=0" delete files /bin/foo
Error (ccc19.40g): malformed command; expected src:dst pairs
```

cmd/minimega/disk.go
```diff
@@ -384,15 +427,24 @@ func cliDisk(ns *Namespace, c *minicli.Command, resp *minicli.Response) error {
...
 		}
 
+		delete := c.BoolArgs["delete"]
+
 		options := fieldsQuoteEscape("\"", c.StringArgs["options"])
 		log.Debug("got options: %v", options)
 
-		pairs, err := parseInjectPairs(c.ListArgs["files"])
+		var files interface{}
+		if len(options) == 0 {
```
jacdavi, line 436: I accidentally found a panic here if you use `disk inject image.qcow2 options "" files /bin/foo` as this if evaluates to true, but the parser has made `files` a string. 

Fix should just be `if _, ok := c.StringArgs["options"]; !ok {` so we just check whether the user specified options

jacdavi (changes requested):
Sorry for so much back and forth. Two final things and then it should be all set!

cmd/minimega/disk.go
glattercj: Fixed

cmd/minimega/disk.go
glattercj: I'm not able to repro this. I changed the `if` check to match your fix but here's what I'm getting both before and after making the change:

```bash
$ disk inject image.qcow2 options "" files /bin/foo
Error (crn1): invalid command: `disk inject image.qcow2 options  files /bin/foo`
```

If I add the `delete` keyword it just returns without any output. Looking at some debug output the minicli.Command just turns the empty quotes into a blank so `c.StringArgs["options"]` returns the next token being `delete`.

```bash
$ disk inject image.qcow2 options "" delete files /bin/foo

```

jacdavi: Hmm it was happening to me every time. Regardless it looks good now

```
minimega[minimega]$ disk inject /scratch/images/test.qcow2 options "" files /bin/foo
panic: runtime error: index out of range [0] with length 0

goroutine 99 [running]:
main.parseFiles({0x8d0cc0?, 0xc0000a4df0?})
        /scratch/minimega/cmd/minimega/disk.go:370 +0x276
main.cliDisk(0x0?, 0xc0000c2780, 0xc00053e400)
        /scratch/minimega/cmd/minimega/disk.go:442 +0x5eb
main.wrapSimpleCLI.func1(0xc0000c2780?, 0xc0001129c0?)
        /scratch/minimega/cmd/minimega/cli.go:104 +0x8b
github.com/sandia-minimega/minimega/v2/pkg/minicli.ProcessCommand.func1()
        /scratch/minimega/pkg/minicli/minicli.go:173 +0x196
created by github.com/sandia-minimega/minimega/v2/pkg/minicli.ProcessCommand
        /scratch/minimega/pkg/minicli/minicli.go:160 +0xca
```

cmd/minimega/disk.go
```diff
@@ -328,6 +355,39 @@ func parseInjectPairs(files []string) (map[string]string, error) {
...
+// minicli.Command StringArg, otherwise it gets turned into a ListArg.
+// this function returns either a paths string slice if 'delete' is part
+// of the original command, or a pairs string map.
+func parseFiles(files interface{}) (map[string]string, []string, error) {
+	var pairs map[string]string
+	var paths []string
+	var err error
+	switch v := files.(type) {
+	case []string:
+		if !(sliceContainsCharacters(v, ",:")) {
```
jacdavi, line 369: Final final thing I promise :) 

With this change we no longer get the error "malformed command; expected src:dst pairs" when running `disk inject /scratch/images/test.qcow2 files /bin/foo`.

Maybe the code should just be explicit whether it expects single files or pairs based on the delete flag?

jacdavi (changes requested):

cmd/minimega/disk.go
glattercj: @jacdavi I changed the logic a bit to handle this case and tried all the previous ones you had asked about. I think it covers them all. Let me know what you think.

glattercj: @jacdavi bump

jacdavi: Thanks for the reminder, looks good now!

jacdavi (approved):

PR #1549, opened 2024-12-02
feat: Create automated debian and RPM packaging

### LINKED ISSUE #1286: misc: add RPM packaging
Mirror the Debian packaging for RPMs.

sdelliot:
This PR provides several needed packaging changes to minimega.

1. A `systemd` script that is used to start minimega as a service and will be packaged with the minimega release files. The old `initd` script is unusable on modern versions of Linux. To automate the installation of the system service on package install, this update was necessary.
2. A build system for compiling minimega for RHEL-based distributions by developing an RPM spec file for minimega. (Closes #1286)
3. Updated the debian build system using more modern and maintainable approaches to building a debian package.
4. Finally, we add github actions to create both the Debian and RPM release files so that they are automatically created when a new release is issued.

To use the new packaging setup, a README file is provided for both debian and rpm-based distributions. The build scripts are designed to work with debian-based systems. To properly install the RPM, users will also need to install/activate the EPEL repo. See [this document](https://www.redhat.com/en/blog/install-epel-linux) for details on how to do this.


misc/daemon/minimega.service
```diff
@@ -0,0 +1,30 @@
...
+Restart=on-success
+# Setting ulimit and environment variables
+LimitNOFILE=1024000
+LimitNPROC=4096000
+EnvironmentFile=/etc/minimega/minimega.conf
+# Pre Start script
+ExecStartPre=/bin/mkdir -p /etc/minimega/saved_vms/
+ExecStartPre=/bin/mkdir -p ${MM_RUN_PATH}
+# Starting minimega
+ExecStart=/usr/bin/minimega -base=${MM_RUN_PATH} -filepath=${MM_FILEPATH} -degree=${MM_MESH_DEGREE} -nostdin=${MM_DAEMON} -context=${MM_CONTEXT} -port=${MM_PORT} -level=${MM_LOG_LEVEL} -logfile=${MM_LOG_FILE} &
```
GhostofGoes, line 23: There are a lot of additional variables that have been added to the Dockerized config that aren't present in the `minimega.conf` used by the init daemon (see [start-minimega.sh](https://github.com/sandia-minimega/minimega/blob/master/docker/start-minimega.sh)). Should those be added to this script, as well as `minimega.conf` and the old init script? 

Some examples of variables: `MM_BROADCAST` for `-broadcast`, `MM_VLANRANGE` for `-vlanrange`, and `MM_APPEND` to append additional arguments that don't have variables (this was added recently in #1542 ).

sdelliot: Adding the new parameters to the `minimega.conf` and `systemd` file definitely makes sense and I'll push that soon. 

Personally, I don't think it makes sense to update the `init.d` script as that system has largely been removed from most popular Linux distributions and I don't know if it's worth maintaining that code. If fact I think it should probably be deprecated in the next release (and subsequently removed the following release) as Docker and `systemd` are better options.

In looking closer at the various config options it looks like there is a naming mismatch between [start-minimega.sh](https://github.com/sandia-minimega/minimega/blob/master/docker/start-minimega.sh) and [minimega.conf](https://github.com/sandia-minimega/minimega/blob/master/misc/daemon/minimega.conf). For example, `MM_RUN_PATH` is passed in as the `-base` parameter with `minimega.conf` and this variable got renamed to `MM_BASE` in `start-minimega.sh`. I don't think that this PR is necessarily the right place to modify those variable names as 1) It feels like scope creep and 2) I suspect any renaming will cause breaking changes to existing users setup. 

One possible solution is to add "duplicate" variables (e.g, `MM_RUN_PATH` and `MM_BASE` to minimega.conf and move the `systemd` file to use those new ones. Then we can deprecate the "old" variable names and remove them when we remove the `init.d` script in a few versions. Thoughts?

sdelliot: I added the new variables (with a systemd preference to use the new version). Let me know if you would like any other changes.

GhostofGoes: The changes look good. I agree with you on all points. 

Regarding the `init.d` scripts, systemd is how most of the major distributions operate nowadays, and there isn't a reason to continue maintaining the old scripts. Better to leave them alone and not risk breaking using still running on older distros.

For duplicate variables, your solution is good, and I don't see any obvious downsides to that approach.

GhostofGoes: Also, just so it's clear this PR will need review by a minimega developer prior to merge. I'm just a humble downstream user with a vested interest in improvements to how it's deployed 😁

jacdavi (changes requested):
Thanks for the PR, this looks great! I am by no means an expert at packaging, but I have a few comments. I have built both images successfully, but probably need to do some more testing

1. Can we standardize where the binaries are placed (root vs `packaging/rpm/rpmbuilds/RPMS`). 
2. When building the RPM, it looks like minimega gets built twice (looks like once in `build.bash` and again in `minimega.spec`. Do we need both?
3. After building I see a change in `minimega.spec` adding the version. Can that be updated to work like `debian/changelog.base`?

sdelliot:
Excellent suggestions! I just pushed some changes and can test locally later today. The one last thing I think we _may_ want to do, though I'll leave it up to you, is modifying the following:

```diff
diff --git a/.github/workflows/deb.yml b/.github/workflows/deb.yml
index c4141fe1..a66b0816 100644
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -1,5 +1,7 @@
 name: DEB Build
-on: push
+on:
+  release:
+    types: [published]
 
 jobs:
   build:
diff --git a/.github/workflows/rpm.yml b/.github/workflows/rpm.yml
index 7d2782b2..6dae5774 100644
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -1,5 +1,7 @@
 name: RPM Build
-on: push
+on:
+  release:
+    types: [published]
 
 jobs:
   build:
```

This will only build the new RPM/DEB only when a new release is published rather than each time a new merge/push occurs. There are pros/cons to each approach where leaving the current state provides up-to-date packages for users between release commits but at a "cost" of more CI time.

sdelliot:
I tested it out and I think all the requested changes have been implemented.

jacdavi:
> Excellent suggestions! I just pushed some changes and can test locally later today. The one last thing I think we _may_ want to do, though I'll leave it up to you, is modifying the following:
> 
> ```diff
> diff --git a/.github/workflows/deb.yml b/.github/workflows/deb.yml
> index c4141fe1..a66b0816 100644
> --- a/.github/workflows/deb.yml
> +++ b/.github/workflows/deb.yml
> @@ -1,5 +1,7 @@
>  name: DEB Build
> -on: push
> +on:
> +  release:
> +    types: [published]
>  
>  jobs:
>    build:
> diff --git a/.github/workflows/rpm.yml b/.github/workflows/rpm.yml
> index 7d2782b2..6dae5774 100644
> --- a/.github/workflows/rpm.yml
> +++ b/.github/workflows/rpm.yml
> @@ -1,5 +1,7 @@
>  name: RPM Build
> -on: push
> +on:
> +  release:
> +    types: [published]
>  
>  jobs:
>    build:
> ```
> 
> This will only build the new RPM/DEB only when a new release is published rather than each time a new merge/push occurs. There are pros/cons to each approach where leaving the current state provides up-to-date packages for users between release commits but at a "cost" of more CI time.

I did think about this originally, but there's currently no restrictions on the docker build, so I'm fine to always build these as well. All three can be addressed together in a future PR if needed

sdelliot:
> Was able to install the rpm without issue. For the deb, I got this error trying to install both on our country club cluster and an ubuntu 22 VM. Any idea what the issue would be?
> 
> ```
> # dpkg -i minimega_2.9_amd64.deb
> (Reading database ... 87624 files and directories currently installed.)
> Preparing to unpack minimega_2.9_amd64.deb ...
> Unpacking minimega (2.9) over (2.9) ...
> Setting up minimega (2.9) ...
> minimega user already exists.
> Could not execute systemctl:  at /usr/bin/deb-systemd-invoke line 142.
> ```

This should be fixed with [c6a89da](https://github.com/sandia-minimega/minimega/pull/1549/commits/c6a89dad99108d43954c545337e11de04ab85b22). Let me know if you have any other issues!

jacdavi:
> > Was able to install the rpm without issue. For the deb, I got this error trying to install both on our country club cluster and an ubuntu 22 VM. Any idea what the issue would be?
> > ```
> > # dpkg -i minimega_2.9_amd64.deb
> > (Reading database ... 87624 files and directories currently installed.)
> > Preparing to unpack minimega_2.9_amd64.deb ...
> > Unpacking minimega (2.9) over (2.9) ...
> > Setting up minimega (2.9) ...
> > minimega user already exists.
> > Could not execute systemctl:  at /usr/bin/deb-systemd-invoke line 142.
> > ```
> 
> This should be fixed with [c6a89da](https://github.com/sandia-minimega/minimega/pull/1549/commits/c6a89dad99108d43954c545337e11de04ab85b22). Let me know if you have any other issues!

Thanks for looking into that. Everything looks good now!

jacdavi (approved):

PR #1555, opened 2025-03-18
refactor: Refactor snapshot/migrate functionality to rename to "save"

sdelliot:
This PR accomplishes the primary task of #1540 with the refactoring of `vm snapshot` and `vm migrate`. This includes the following:

**Breaking Changes**
* Rename `vm migrate`  -> `vm save`
* Rename `ns snapshot`  -> `ns save`
* Rename `vm configure migration` -> `vm configure state`

**Additional Changes**
* Refactoring the code to replace most references to migrate/snapshot. There are some exceptions including 1) where snapshot is used as in `disk snapshot` and 2) where there were only internal references but could conflict with "state" where state could be overloaded (e.g., `vm state` (BUILDING, RUNNING,...) or the saved VM memory).
* Fixed multiple bugs where the image that is "restored" would not work properly if there was a backing image. It also did not account for extra disk parameters (e.g., "virtio,writeback"
* Adding namespace queueing to the `ns save` output.


activeshadow:
@sdelliot is it possible to leave the old commands in place for some amount of time (perhaps marked as deprecated)? Renaming the commands will break some functionality in phenix, so if they can't be left in for a little bit then we might want to coordinate some activities.

Thoughts?

/cc @jacdavi @glattercj

sdelliot:
> @sdelliot is it possible to leave the old commands in place for some amount of time (perhaps marked as deprecated)? Renaming the commands will break some functionality in phenix, so if they can't be left in for a little bit then we might want to coordinate some activities.

That totally makes sense to me. I adjusted it so that the old commands still exist but point to the new functionality. So there should no longer be API changes but the underlying functionality will still be modified. I tested it out and it looks like it works as expected.

cmd/minimega/disk.go
```diff
@@ -113,6 +113,25 @@ func diskSnapshot(src, dst string) error {
...
+
+		// Now we can copy the backing file to the destination
+		out, err := processWrapper("cp", src, dst)
+		if err != nil {
+			return fmt.Errorf("[image %s] error copying to destination: %v: %v", src, out, err)
+		}
+		log.Info("Copied existing snapshot file from %s to %s", src, dst)
+		return nil
+	}
+
```
jacdavi, line 134: I'm a bit confused by this section, but may just be getting things mixed up.

If a user has disk B backed by disk A and calls "snapshot B C" my expectation would be a backing chain of `C -> B -> A`, but it seems like this will produce `C(really A),  B -> A` which breaks the chain. Is there a reason we want that?

jacdavi (commented):
Haven't had a chance to run things yet, but did a quick read through.

One general question, compared to #1540 we're losing the functionality of saving just the disk of a VM. I think that's what we had agreed on previously, but I'm just wondering again if that feature was useful.

cmd/minimega/disk.go
sdelliot: So this section was largely put in for two reasons. 
1. The chain of snapshots `C -> B -> A` was not working correctly. I haven't tested now that #1557 was merged, but that may have fixed the issue. It also could have been that one of the intermediate images, `B` wasn't in the "files" directory. I'll test with the latest changes and see what happens.
2. The secondary reason is that having a chain of snapshots makes it more difficult for a user to actually capture (save off/compress/etc.) the entire chain. Suppose you have a chain `B->A` where `A` is the backing image and `B` is the running version. Then you snapshot that, we save off the state of `B`. Then when you start the experiment again, this method is called and creates a new snapshot `C` which has the chain `C->B->A`. With this code, rather than creating a new `C` we will take a copy of `B`. So that if you snapshot again the original case would end up with `C->B->A` , but this version will end up with `B'->A` where `B'` is `C+B`. Basically, instead of creating continually longer chains, you never end up with a chain larger than two.

sdelliot:
> One general question, compared to #1540 we're losing the functionality of saving just the disk of a VM. I think that's what we had agreed on previously, but I'm just wondering again if that feature was useful.

That's correct. How it was previously implemented, I believe the VM has to be paused to save off the disk, assuming that's correct, there is no loss of functionality as users can just ignore the memory state. If we don't pause a VM and instead used a live migration of just the disk then there would be some loss of functionality.

cmd/minimega/disk.go
sdelliot: Thanks for the reminder! I just tested with the new logic. It does work much better, but still needed an additional bug fix due to a change that I think was introduced in #1545. It now works as it originally did. 

As for the second point, maybe that should be handled in a different PR that adds some kind of "export" functionality. The point was to make it easier to actually transport experiments from one system to another. A chain of snapshots makes that much more fragile as the user would need to collect all of them and then rebase those paths once they are on the new system. Regardless, this is beyond the scope of this PR.

jacdavi: Ok thanks! I agree that what you had is probably the better option 90% of the time and another PR for it makes sense. I just don't want the meaning of the `disk` commands to differ too much from the underlying QEMU versions.

Will do a final review and merge this week!

jacdavi (approved):

PR #1556, opened 2025-04-01
update minimum go version to 1.20; update crypto/net/sys packages

jacdavi:
This handles some dependabot alerts that didn't include the go update


aherna (approved):
Lgtm

PR #1557, opened 2025-04-02
fix error when rebasing onto no backing image

jacdavi:
Small fix for recent PR #1545 

During development I changed rebasing to make the new image relative to the first. However, in doing so I didn't double check that a rebase onto "" (e.g., make the image independent) still worked. This PR fixes that issue


aherna (approved):
Lgtm

PR #1565, opened 2025-06-25
feat: add actions for building and creating a release; bump version to 3.0.0

jacdavi:
This PR makes the following changes:

* Adds a new `build` action which simply runs `./scripts/all.bash` to verify basic formatting and compilation on every push.
  * This aids in reviewing PRs and ensures a passing build. It is also used in the `release` action
* Adds a new `release` action which is manually triggered. When run it will create a draft release in GitHub with 1) automated release notes from commit messages and 2) attached binaries/deb/rpm files.
  * This will make it possible to create new releases in only a few minutes
* Bumps version to 3.0.0
  * The reason for bumping to 3.0.0 is to start conforming to conventional commits. A recent PR (#1555) introduced breaking changes necessitating a major release


sdelliot (approved):

sdelliot:
> @sdelliot do you think we can move forward with merging this and create a separate issue for python packaging?
> 
> Motivation for asking now is that someone reached out that miniweb directories are broken in the 2.9 deb, and a new release fixes that due to your packaging improvements.

We can save the python packaging for a separate issue. Will review/merge shortly.

PR #1567, opened 2025-07-22
build(deps): bump jquery to 3.7.1

wmstout:
Bump jQuery from 3.1.1 to 3.7.1 due to multiple vulnerabilities associated with versions <3.5


jacdavi:
I agree that we should remove the old versions.

Also worth noting that `doc/content/static/jquery.js` and `doc/content/static/mega.js` are both jquery `1.8.2`. I'm fine not addressing those in this PR, but they may also cause failing scans

GhostofGoes (approved):

PR #1570, opened 2025-08-08
Update disk.go with more attempts to find partitions

albuwab:
In larger experiments (300+ VMs), this action fails to find partitions on the image. I believe increasing the number of possibles may patch this issue while investigate root causes. Performance issues may be hindering the ability for large experiments to find these objects and cause it to fail. Increasing this count will allow greater chance for success. Better handling may improve the experience instead of failing out of an experiment launch entirely.


jacdavi (approved):
This seems reasonable. The timeout only applies when there are issues finding paritions, so extending the retry time shouldn't affect most situations.

Issue #1572, opened 2025-08-22
Bug: tool binaries missing from Docker image


GhostofGoes:
**Describe your environment**

1. minimega version: minimega 2.9 e8ce6d1f343be57f1a5739e9af8477dbaf08ba5d 2025-07-29
2. Linux distro/version: Ubuntu 22.04.5 LTS
3. Go compiler version: 1.20.14
4. VM types: N/A

**Describe the bug**
There are a number of tools listed [in the documentation](https://www.sandia.gov/minimega/tools/), such as `rfbplay`, `vncdrone`, `vmbetter`, etc. However, these tools are not included in the Docker image. This resulted in being unable to use `rfbplay` to replay a framebuffer recording captured with `vnc record`. Additionally, `ffmpeg` is missing, and might be worth including in the image.

8 binaries in `/opt/minimega/bin/` in final image, vs 26 binaries in `/minimega/bin/` directory as build outputs. 18 binaries NOT being included. `bin/` is 67 MB in current minimega image, vs 136MB in build output. That's a 69MB difference, which is relatively small.

I'll be opening a PR shortly with the fix for the missing tools, but wanted to document in an issue.

**To Reproduce**
1. `minimega -e vnc record fb vm-name /phenix/test.fb`
2. `minimega -e vnc stop fb vm-name`
3. `rfbplay /phenix/test.vnc /phenix/test.mp4`

**Expected behavior**
`/phenix/test.mp4` file is generated.

PR #1573, opened 2025-08-22
fix: include tools in docker image

GhostofGoes:
This PR includes the tool binaries in the docker image.

While this includes some tools like "minitest" that usually aren't needed in a deployment, the disk utilization difference is minimal, and this ensures any future tools get included in the Docker image. It also simplifies the Dockerfile a bit (less commands and potentially less layers).

8 binaries in `/opt/minimega/bin/` in final image, vs 26 binaries in `/minimega/bin/` directory as build outputs. 18 binaries NOT being included. `bin/` is 67 MB in current minimega image, vs 136MB in build output. That's a 69MB difference, which is relatively small.

`rfbplay` requires `ffmpeg`, which increases the image size by a bit (230MB or so), so I'm making a separate PR for adding ffmpeg to elicit debate about if that's worth adding.

Image without binaries: 774MB
Image with binaries: 840MB
Image with ffmpeg and binaries: 1.07GB

Relevant issue: #1572

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,31 +31,22 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/* \
   && rm -rf /var/cache/apt/archives/*
 
-COPY --from=gobuilder /minimega/bin/minimega /opt/minimega/bin/minimega
-COPY --from=gobuilder /minimega/bin/miniweb  /opt/minimega/bin/miniweb
-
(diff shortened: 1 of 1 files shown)


jacdavi (approved):
Sorry I forgot to get to this PR. 

This looks good to me, and I definitely agree those tools should be included. Since we only have 1 docker image, it should be able to do everything advertised imo. If someone has a space constraint in the future, I think we could create a separate `-slim` image and make reductions there

PR #1579, opened 2025-09-17
fix: miniccc client connection issues

nryan100:
This handles client initialization errors before we attempt to read from the client connection. 
Discovered this issue when using this command to start the miniccc client: `/miniccc -family unix -parent /cc -logfile /miniccc.log &`
For certain cases, the socket file does not exist when we try to connect to it and this error is thrown:
image
As the client could not connect to the socket, attempting to read from that client throws this nil pointer error. Checking for any errors before we attempt a read resolves this.


glattercj (approved):
LGTM

PR #1581, opened 2025-09-19
fix: Remove tailing space in comment

nryan100:
Fix tailing space in comment (from my PR #1579)  breaks the `gofmt` check in scripts/check.bash


jacdavi (approved):
Appreciate the fix, and my bad for not catching it! #1565 will do a build on PRs making this easier to catch

PR #1584, opened 2025-10-16
feat: add option for using snapshot for first disk only

activeshadow:
This adds a new `vm config snapshot-first-disk-only` (I know, that's ugly) option that will do just what it says - only snapshot the first disk provided. This way any other disks will be persistent, allowing the best of both worlds (boot from snapshot, but mount and persist data from within the VM).

I need someone to look at the diff to `vmconfiger_cli.go`, as I think some of the previous commits that were updating snapshot functionality may have edited this file directly and I want to make sure the changes don't cause any regressions. This file is actually auto generated using the `vmconfiger` executable via `go generate` using the `BaseConfig, KVMConfig, and ContainerConfig` types as the source.

/cc @jacdavi @sdelliot @glattercj


GhostofGoes:
What about an option to select which disk(s) should be snapshotted?

GhostofGoes:
Any plans to finish this @activeshadow ? the vmconfiger output will probably need to be regenerated after rebasing.

activeshadow:
Wow, holy crap it's already almost been a year. :joy:

Yes @GhostofGoes I'd like to finish this. I'll take a look again ASAP. Still may need a few others to help me review.

GhostofGoes:
Absolutely, happy to review once it's ready!

PR #1585, opened 2025-10-31
fix(minirouter): update bird template to work with bird v2 and v3

activeshadow:
Starting with bird v2 (at least from what I can tell), the config file changed such that each protocol supported channels. This allowed a single bird executable to work for both IPv4 and IPv6 (among other things).

Note that this came about because:

1. minirouter can no longer run on Ubuntu Noble due to outdated glibc
2. updating to a more recent distro no longer has the bird v1 package
3. installing bird v2 or bird v3 leads to bird failing to start due to config file issues
4. installing bird v2 or bird v3 also causes an error due to the bird6 command missing

As written, this fix supports IPv4 and IPv6 for static routes (named and unnamed), but only supports IPv4 for OSPF and BGP.

> [!WARNING]
> THIS NEEDS TESTING! I'm only using bird for unnamed static routes right now, so I have not tested that this works properly for named static routes, OSPF, or BGP via bird. As such, I will try to test when I can unless someone else has time to test.

At a minimum, this will allow bird to start properly. Note that bird not starting because of the invalid config file does not prevent minirouter from running properly.

### DIFF
--- a/cmd/minirouter/bird.go
+++ b/cmd/minirouter/bird.go
@@ -321,7 +321,7 @@ protocol kernel kernel_ipv4 {
     export all;   # Actually insert routes into the kernel routing table
   };
 
-	learn;
+  learn;
 }
 
 protocol kernel kernel_ipv6 {
@@ -332,7 +332,7 @@ protocol kernel kernel_ipv6 {
     export all;   # Actually insert routes into the kernel routing table
   };
 
-	learn;
+  learn;
 }
 
 # The Device protocol is not a real routing protocol. It doesn't generate any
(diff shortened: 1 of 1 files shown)


GhostofGoes:
@activeshadow Has this been tested? Any plans to merge this eventually?

activeshadow:
> @activeshadow Has this been tested? Any plans to merge this eventually?

@GhostofGoes tested, no. Plans to merge eventually, yes.

activeshadow:
@GhostofGoes I still plan on getting this one done too.

PR #1595, opened 2026-02-20
chore: Bump to go 1.24 and update vendored packages

sdelliot:
## Description ##
This update bumps the minimum go version to 1.24 and updates all vendored packages. Additionally, we correct several issues caused by `go vet` with 1.24. Aside from the needed `go vet` changes, standard golang commands were used (`go get -u`, `go mod tidy`, `go mod download`, `go mod vendor`.

## Motivation and context ##

This enables a more modern workflow for minimega and makes progress towards updating various vendored dependencies, which contain numerous bug fixes.

This PR consolidates related dependabot PRs including: #1586 and #1559 

## Testing ##

After building the new version this was manually tested an an experiment was successfully launched. All minimega test cases continue to pass.

## Checklist ##


doc/content/articles/installing.article
```diff
@@ -44,10 +44,10 @@ Having installed the dependencies, grab the minimega source:
        $ git clone git@github.com:sandia-minimega/minimega.git
        $ cd minimega
 
-Next, check out the 2.7 release. If you wish to run the development version
+Next, check out the latest release. If you wish to run the development version
 ("tip") of minimega, skip this command.
 
-	$ git checkout 2.7
+	$ git checkout 3.0.0
```
jacdavi: Wonder if it's worth going a step forward to generalize this with something like

```
git fetch --tags
git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
```

Should checkout the latest tagged release, but maybe a bit overkill here

jacdavi (approved):
Looks good. I know some people in the past have had issues because the system they run minimega on can't easily upgrade golang. So I think we can leave this open for a bit.

Also curious, would bumping the golang version be considered a breaking change? It's probably not in the traditional sense, but it will break some environments from compiling without further changes.

sdelliot:
> Looks good. I know some people in the past have had issues because the system they run minimega on can't easily upgrade golang. So I think we can leave this open for a bit.

I'm not a golang expert, but I thought that this would only impact those that need to _compile_ the binaries. That is, if users use the pre-compiled versions, it should still "just work"? Are there examples of where this would not be the case?

> Also curious, would bumping the golang version be considered a breaking change? It's probably not in the traditional sense, but it will break some environments from compiling without further changes.

Personally, I don't think this is a breaking change. I did some searching and it seems that this is generally the consensus as well due to golang's strong backwards compatibility. Happy to hear others chime in as well though.

doc/content/articles/installing.article
sdelliot:
```suggestion
	$ git fetch --tags
	$ git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
```

jacdavi:
@sdelliot I believe we got the go ahead to merge this, so whenever you're ready feel free to merge.

GhostofGoes:
@sdelliot Dockerfile needs to be updated

GhostofGoes (changes requested):
Update Dockerfile and verify the Docker build is still working.

nblair2:
> I'm not a golang expert, but I thought that this would only impact those that need to _compile_ the binaries. That is, if users use the pre-compiled versions, it should still "just work"? Are there examples of where this would not be the case?

Compiled go should just work, you can run into problems with the libc compatibility (ref #1597)

sdelliot:
> Update Dockerfile and verify the Docker build is still working.

Docker file is updated in [3cc7de7](https://github.com/sandia-minimega/minimega/pull/1595/commits/3cc7de7bae1a2f39683e702aa33d43473b32cd58) and I confirmed that it still works as expected.

GhostofGoes (approved):

PR #1597, opened 2026-03-11
build(packages): staticly link binaries

nblair2:
## Description ##
Statically link packages with `CGO_ENABLED=0`.

## Motivation and context ##
Allows packages to run on hosts without any dependencies.

## Testing ##
TODO

## Checklist ##
- [X] This PR conforms to the process detailed in the [Contributing Guide](https://github.com/sandia-minimega/minimega/blob/master/.github/CONTRIBUTING.md).  
- [X] I have included no proprietary/sensitive information in my code. 
- [X] I have performed a self-review of my code.
- ~I have commented my code, particularly in hard-to-understand areas.~
- [ ] I have tested my code using the methods described above.

## Additional Notes ##
N/A

### DIFF
--- a/scripts/build.bash
+++ b/scripts/build.bash
@@ -40,7 +40,11 @@ echo "BUILD PACKAGES (linux)"
 for i in ${DIRECTORY_ARRAY[@]}; do
     for j in `ls $i | grep -v plumbing`; do
         echo $j
-        go install -ldflags="-s -w" $i/$j
+        if [[ "$j" == "miniccc" || "$j" == "protonuke" || "$j" == "minirouter" ]]; then
+            CGO_ENABLED=0 go install -ldflags="-s -w" $i/$j
+        else
+            go install -ldflags="-s -w" $i/$j
+        fi
         if [[ $? != 0 ]]; then
             exit 1
         fi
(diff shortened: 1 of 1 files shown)


scripts/build.bash
```diff
@@ -40,7 +40,11 @@ echo "BUILD PACKAGES (linux)"
 for i in ${DIRECTORY_ARRAY[@]}; do
     for j in `ls $i | grep -v plumbing`; do
         echo $j
-        go install -ldflags="-s -w" $i/$j
+        if [[ "$j" == "miniccc" || "$j" == "protonuke" || "$j" == "minirouter" ]]; then
```
GhostofGoes, line 43: Make this a variable that can be changed?

nblair2: @GhostofGoes I have not had a chance to test any of this yet, which is why it is still in draft.

My main goal is to compile miniccc without `CGO` so that we don't need tight coupling between a miniccc version and a host's version of GLIBC. For example, miniccc present in a modern minimega container will not run on an older OS like Ubuntu 18.04.

I would prefer this to be the default, otherwise its still a manual process to recompile myself. However, I don't have a good grasp of downstream changes for other use cases so defer to you on if this makes sense or not.

PR #1603, opened 2026-06-07
fix(qemu-nbd): claim device on check

nblair2:
## Description ##
Serialize NBD device selection and connection behind a package mutex held across `GetDevice` and `qemu-nbd -c`, and retry the next device if the connect fails.

## Motivation and context ##
`GetDevice` reports a device as free, but `ConnectImage` only connects it afterwards, with nothing reserving it in between. Two concurrent disk operations can therefore select the same device and one `qemu-nbd -c` fails. Holding the lock through the connect (the point at which the device gains a pid/lock file) closes that window; retrying on failure also covers a lost race against an external user.

## Testing ##
Tested in the same heavy-use scenario where write contention was encountered - no issue.

## Checklist ##
- [X] This PR conforms to the process detailed in the [Contributing Guide](https://github.com/sandia-minimega/minimega/blob/master/.github/CONTRIBUTING.md).
- [X] I have included no proprietary/sensitive information in my code.
- [X] I have performed a self-review of my code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have tested my code using the methods described above.
- [X] All GitHub Actions are passing.

## Additional Notes ##
N/A

### DIFF
--- a/internal/nbd/nbd.go
+++ b/internal/nbd/nbd.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/sandia-minimega/minimega/v2/pkg/minilog"
@@ -18,6 +19,10 @@ import (
 
 var (
 	ErrNoDeviceAvailable = errors.New("no available nbds found")
+
+	// nbdLock serializes device selection and connection so concurrent
+	// callers can't claim the same device.
+	nbdLock sync.Mutex
 )
 
 const (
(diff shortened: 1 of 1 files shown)


GhostofGoes (approved):

GhostofGoes:
@cmulk could use your eyes on this since you're battle-hardened with `nbd`

cmulk:
This makes sense to me, I'm good with it, thanks!

glattercj (approved):

PR #1614, opened 2026-07-22
feat(vmconfig): add initial Android VM configuration plumbing

sdelliot:
## Description ##
This PR adds the initial Android VM configuration plumbing needed for **future** native Android runtime support in minimega.

The change is intentionally limited to configuration and validation support. It does **not** introduce a new Android VM runtime type or Android lifecycle management yet. This capability has be prototyped using minimega/FIREWHEEL/android studio, but we plan on using a phased approach for slowly introducing changes to minimega that should not live in FIREWHEEL or random hacky scripts.

Specifically, this PR:
- adds `AndroidConfig` to the VM configuration model
- integrates Android config into `VMConfig` copy/string/read/write/clear flows
- Updates `check` so that there are optional Android runtime dependency validation and ensures that Android validation is only performed when Android appears configured
- updates the KVM `MigratePath` config field to preserve the canonical `state` name while continuing to accept the deprecated `migrate` alias in generated output

## Motivation and context ##
This change is required to support the first phase of Android integration in minimega.

This work is intentionally scoped to configuration plumbing only. It prepares minimega for future Android runtime support without introducing process lifecycle, networking, or VM type changes yet.

## Testing ##
Testing was performed in two parts:

### Static verification
- Regenerated the generated config handler file:
  ```bash
  go generate ./cmd/minimega
  ```
- Ran:
  ```bash
  go vet ./...
  ```

### Manual config testing
The following configuration flows were exercised successfully:

```text
vm config android-sdk /opt/android-sdk
vm config android-emulator emulator
vm config android-adb adb
vm config android-avd Pixel_9a
vm config android-avd-dir /var/lib/android/avd
vm config android-no-window false
vm config android-extra-args "-prop foo=bar baz" -gpu off
vm config
clear vm config android-avd
```

## Checklist ##


GhostofGoes:
please squash commits

cmd/minimega/android_support.go
```diff
@@ -0,0 +1,183 @@
...
+func androidConfiguredAny() bool {
+	for _, ns := range namespaces {
+		if androidConfigured(ns.vmConfig.AndroidConfig) {
+			return true
+		}
+	}
+	return false
+}
+
+func findAndroidEmulator(path string) (string, error) {
```
GhostofGoes: findAndroidEmulator and findADB have nearly identical logic with exception of hardcoded search string, should be generalized.

cmd/minimega/android_support.go
```diff
@@ -0,0 +1,183 @@
...
+			return "", err
+		}
+		return path, nil
+	}
+
+	return exec.LookPath("adb")
+}
+
+func validateAndroidConfig(cfg AndroidConfig) error {
+	if cfg.ConsoleBasePort > 0 && cfg.ConsoleBasePort < 1024 {
```
GhostofGoes: are negative base ports valid?

cmd/minimega/android_support.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if _, err := findAndroidEmulator(cfg.EmulatorPath); err != nil {
+		return fmt.Errorf("android emulator not found: %v", err)
+	}
+
+	if _, err := findADB(cfg.ADBPath); err != nil {
+		return fmt.Errorf("android adb not found: %v", err)
+	}
+
+	if cfg.RequireKVM && !lsModule("kvm") {
```
GhostofGoes: in what cases is KVM not required? Would be good to document this in the config option. or maybe it's obvious.

cmd/minimega/android_support.go
```diff
@@ -0,0 +1,183 @@
+package main
```
GhostofGoes, line 1: There should be some basic Go tests for this module. Don't need to use minitest.

cmd/minimega/external.go
```diff
@@ -113,6 +113,24 @@ func checkExternal() error {
...
+			cfg := ns.vmConfig.AndroidConfig
+			if !androidConfigured(cfg) {
+				continue
+			}
+
+			if err := checkAndroidDependencies(cfg); err != nil {
+				log.Warn("android runtime unavailable: %v", err)
+				break
+			} else {
+				log.Info("android runtime available")
```
GhostofGoes, line 128: should namespace be logged here?

GhostofGoes (changes requested):

cmd/minimega/android_support.go
sdelliot: I broke up all the Android PRs so that they would be easier to review rather than one huge one. So the basic unittests for this are in #1616.

cmd/minimega/android_support.go
sdelliot: They are now combined

cmd/minimega/android_support.go
sdelliot: Nope, fixed :)

cmd/minimega/external.go
sdelliot: I don't think so? In this case it's more about what is available on the system rather than per-namespace. That is, I don't think we can have a case where two namespaces on the same system have different file system permissions (access to the emulator or the KVM kernel module). I am happy to add that in if my assumption is wrong though.

cmd/minimega/android_support.go
sdelliot: I went ahead and removed that config option as it doesn't really make sense.

GhostofGoes (approved):

PR #1619, opened 2026-07-26
fix(ron): track Once command delivery per client

nblair2:
`Once` delivery was gated on a single global `Command.Sent`, set when the command was copied into an outgoing message rather than when a client actually received it.

Any send would burn it for everyone:

- a handshake catch-up send (`sendCommands(c.UUID)`) for a client the command's `Filter` did not match
- a broadcast that skipped a client not yet in `s.clients`
- a write that failed

The client the `Filter` named then never received the command, and nothing ever reset the flag. `sendCommands` sets `Sent` at message-construction time, but `route` decides delivery per client afterwards — so inclusion in a message never implied delivery to the intended target.

Record delivery per client instead (`Server.onceSent`), and only once the write succeeds. Records outlive the connection, so a reconnecting VM still does not re-run a `Once` command, and are freed when the command is deleted or the command list is cleared (IDs are reused after a reset).

`Command.Sent` is retained but unused so the gob type sent to `miniccc` agents is unchanged.

### Tests

`internal/ron/once_test.go` covers both legs plus the intent guard. Against the parent commit:

```
--- FAIL: TestOnceNotConsumedByOtherClientHandshake
    target never received Once command 1 (got []) -- consumed by a send to a
    client its filter did not match
--- FAIL: TestOnceDeliveredToLateJoiner
    late joiner never received Once command 1 (got [])
```

Both pass with this change. `TestOnceNotResentAfterReconnect` passes before and after, confirming the documented "not re-sent after client reconnections" behaviour is preserved.

### Not addressed here

`c.maxCommandID` is still advanced before `sendMessage`, so a failed write can permanently drop a *non*-`Once` command. Left for a separate change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/ron/command.go
+++ b/internal/ron/command.go
@@ -67,8 +67,11 @@ type Command struct {
 	// once, or if it should be sent after client reconnections.
 	Once bool
 
-	// Sent tracks whether or not this command has been sent already. Only used
-	// when Once is enabled.
+	// Sent is deprecated and unused. Once delivery is tracked per client by
+	// the server (Server.onceSent); a single global bool could be consumed by
+	// a send that never reached the client the Filter named, permanently
+	// suppressing the command for its intended target. The field is retained
+	// so the gob type transmitted to miniccc agents is unchanged.
 	Sent bool
 
 	// plumber connections
--- a/internal/ron/once_test.go
+++ b/internal/ron/once_test.go
@@ -1,0 +1,305 @@
+package ron
+
+import (
+	"encoding/gob"
+	"io"
+	"net"
+	"path/filepath"
+	"testing"
+	"time"
+)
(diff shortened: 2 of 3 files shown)


internal/ron/command.go
```diff
@@ -67,8 +67,11 @@ type Command struct {
 	// once, or if it should be sent after client reconnections.
 	Once bool
 
-	// Sent tracks whether or not this command has been sent already. Only used
-	// when Once is enabled.
+	// Sent is deprecated and unused. Once delivery is tracked per client by
```
GhostofGoes, line 70: I don't understand the reason to not just remove unused flag, can you elaborate?

PR #1623, opened 2026-07-29
fix(qemu): fix AHCI disk and hugepage arguments

GhostofGoes:
## Summary

Fix two issues I found while evaluating if minimega could use latest QEMU release.

- use `ide-hd` instead of the removed `ide-drive` device for AHCI disks
- use QEMU's supported `-mem-path` option for hugepage-backed guest memory, instead of the `-mem-path` option, which as far as I can tell never existed.

## Why

`ide-drive` was [deprecated in QEMU 4.2](https://qemu.weilnetz.de/doc/5.2/system/deprecated.html#ide-drive-since-4-2) and removed in QEMU 6.0, preventing AHCI-attached disks from launching on current QEMU versions. 

The hugepage path was passed with the unsupported `-mem-info` option instead of `-mem-path`.

## Validation

Tests pass. I don't have a way to test these codepaths directly on a running experiment.

### DIFF
--- a/cmd/minimega/kvm.go
+++ b/cmd/minimega/kvm.go
@@ -1425,7 +1425,10 @@ func (vm VMConfig) qemuArgs(id int, vmPath string) []string {
 			}
 
 			args = append(args, "-device")
-			args = append(args, fmt.Sprintf("ide-drive,drive=ahci-drive-%v,bus=ahci.%v", ahciBusSlot, ahciBusSlot))
+			// "ide-drive" was removed in QEMU 6.0; "ide-hd" is the modern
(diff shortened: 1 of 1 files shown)


GhostofGoes:
Related: should we change the minimum supported QEMU version to 6.0?

GhostofGoes:
ping @sdelliot

sdelliot (approved):
This looks good to me. IMO, I don't think we need to mandate a QEMU 6.0 minimum as both `mem-path` and `ide-hd` were added in QEMU pre v1.0 so we shouldn't have any risk of breaking backwards compatibility. Whether there are other reasons (or arguments) for updating a minimum version is an open question for a different issue/PR.

PR #1624, opened 2026-07-29
feat(minimega): add tab completion support for bash, zsh, and fish

GhostofGoes:
## Summary

Adds the ability to generate a bash tab-completion script for `minimega`, similar to `phenix completion bash` in sceptre-phenix.

### `minimega -completion`

A new `-completion` flag prints a bash-completion script to stdout. The generated script:
- Completes minimega's own command-line flags
- Completes minicli commands typed after `-e`/`-attach` (e.g. `minimega -e vm sta<TAB>` → `vm start`)
- Falls back to filename completion otherwise.

Enable it for the current shell session:
```bash
source <(minimega -completion)
```
Or install it persistently:
```bash
minimega -completion | sudo tee /etc/bash_completion.d/minimega >/dev/null
```

This is documented in `doc/content/articles/usage.article`.

### `docker/minimega-wrapper.sh`

Adds a host-side wrapper script (based on sceptre-phenix's `phenix-wrapper.sh`) that transparently forwards `minimega` commands to the `minimega` Docker container via `docker exec`, only allocating a pseudo-TTY when stdout is actually a terminal.

`docker/README.md` documents installing it as `/usr/local/bin/minimega` and using it to enable bash tab completion for a dockerized minimega deployment.

### `docker/Dockerfile` fix

Also includes a small fix: adds the `less` package to the minimega Docker image's apt install list. `less` is required by `minipager` to page command output when running on shell via `-attach`, for example running `help`;.

without `less` installed minimega logs the following.
```
ERROR pager.go:64: problem paging: exec: "less": executable file not found in $PATH
```

This is because minimega hard-codes `less` as the fallback if the `$PAGER` variable is unset or empty. Maybe there's a better pager, but this one works fine and is easy to install.

## Testing

These changes have been tested manually, including the wrapper script, and running minimega with `-attach`. 

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


cmd/minimega/completion.go
```diff
@@ -0,0 +1,97 @@
...
+#
+# Generated by 'minimega -completion'. To enable minimega tab completion for
+# the current shell session:
+#
+#   source <(minimega -completion)
+#
+# To enable it for every session, write the script somewhere bash-completion
+# will find it, e.g.:
+#
+#   minimega -completion | sudo tee /etc/bash_completion.d/minimega >/dev/null
```
sdelliot, line 33: This is the legacy location for bash completions, the currently recommended way is something like ``/usr/share/bash-completion/completions/minimega`` see https://unix.stackexchange.com/a/717732. I think we can leave the legacy example but add the newer path as well. Maybe something like:


```suggestion
#   minimega -completion | sudo tee /usr/share/bash-completion/completions/minimega >/dev/null
#
#  or
#
#   minimega -completion | sudo tee /etc/bash_completion.d/minimega >/dev/null
```

The documentation below will also need to be modified.

sdelliot (changes requested):
I think this is very helpful! However, I have a few questions:
1. Is there a compelling reason to do it like this rather than install it when the minimega package is installed? It may make sense to have both options, for users compiling from source, but there would likely be saved overhead by just doing it for the user.
2. Any thoughts on making this more flexible so that users that prefer `zsh` or some other shell might be able to extend it? Maybe something like `minimega -completion zsh` or `minimega -completion bash`.
3. This thought probably shouldn't be solved by this PR, but one of my biggest issues with using minimega on the CLI is that I use a non-standard "base", so every time I type a command I must include the `-base` flag (and path). This is in spite of having the `MM_BASE` variable defined in /etc/minimega/minimega.conf. The only reason I bring that up is that this bash completion is likely to go unused by users with custom paths, until minimega can correctly guess it's environment.

GhostofGoes:
> I think this is very helpful! However, I have a few questions:
> 
>     1. Is there a compelling reason to do it like this rather than install it when the minimega package is installed? It may make sense to have both options, for users compiling from source, but there would likely be saved overhead by just doing it for the user.
> 
>     2. Any thoughts on making this more flexible so that users that prefer `zsh` or some other shell might be able to extend it? Maybe something like `minimega -completion zsh` or `minimega -completion bash`.
> 
>     3. This thought probably shouldn't be solved by this PR, but one of my biggest issues with using minimega on the CLI is that I use a non-standard "base", so every time I type a command I must include the `-base` flag (and path). This is in spite of having the `MM_BASE` variable defined in /etc/minimega/minimega.conf. The only reason I bring that up is that this bash completion is likely to go unused by users with custom paths, until minimega can correctly guess it's environment.

1. Our use case is minimega in a container, and thus there is no "install" on the host. This enables the completion script to be called from `.bashrc` or such.
2. +1, I'll add support for other shells.
3. This isn't something I'm familiar with, but my intuition is the CLI *should* read `MM_BASE` from env to alleviate this. Thoughts?

GhostofGoes:
Addressed review comments. Also added `fish` completion for @cmulk because I think he would appreciate it. Fish completion is completely untested, will rely on users to submit any issues.

sdelliot:
>     1. Our use case is minimega in a container, and thus there is no "install" on the host. This enables the completion script to be called from `.bashrc` or such.

That's a compelling enough reason for me!

>     2. +1, I'll add support for other shells.

Thank you!

>     3. This isn't something I'm familiar with, but my intuition is the CLI _should_ read `MM_BASE` from env to alleviate this. Thoughts?

I completely agree! I will go ahead and open up an issue for this. We can ignore that for this PR though.

sdelliot (approved):
This all works for me!

PR #1629, opened 2026-08-05
chore(ci): add tests workflow for GitHub Actions

GhostofGoes:
## Description ##
Adds a GitHub Actions workflow (`.github/workflows/test.yml`) that runs `scripts/test.bash` on `ubuntu-latest`.

## Motivation and context ##
minimega had no CI workflow executing its test suite. This adds automated test coverage on push/PR, matching the pattern of the existing `build.yml`/`deb.yml`/`rpm.yml` workflows.

## Testing ##
The workflow passes wheeee.

## Checklist ##

## Additional Notes ##
Considering all the other workflows, I'm rather surprised tests were never being run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,26 @@
+name: Test
+on:
+  push:
+  pull_request:
+  workflow_call:
+
+jobs:
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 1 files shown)


nblair2 (approved):
Is there any reason the script only runs 'go test', with no args ('-v', '-race') ?

I don't think GitHub has any test integrations with formats like cobertura or junit so no need to do those conversions.

GhostofGoes:
> Is there any reason the script only runs 'go test', with no args ('-v', '-race') ?
> 
> I don't think GitHub has any test integrations with formats like cobertura or junit so no need to do those conversions.

Not sure. I'm still relatively new to Golang, still learning the tools. test.bash has existed for a long time. A separate PR to modernize it would be welcome ;)

PR #1636, opened 2026-08-27
docs(training): update minidoc training modules

### LINKED ISSUE #1407: minidoc: training document changes and updates
***Module 1
- Slide 16 – Don’t need OSPF for router sandwich because the router serves as the default gateway to both networks and it knows how to get everywhere.
- Should use router router dhcp 10.0.0.0 router 10.0.0.1 and router router dhcp 20.0.0.0 router 20.0.0.1 instead of ospf. This sets default gateway as part of dhcp process for vms

***Module 2.5
- Should probably rename this as its own module like module 3 and increment the others
- Slide 5 should talk two how the containers are implemented in minimega as they are different than using docker or LXC linux containers

***Module 3
- Slide 22 link to module 4 title doesn’t match module miniweb vs visualization

***Module 4
- Slide 5 consider changing first bullet to say if you like the minimega console to be available in the web application…

***Module 5
- LGTM

***Module 6
- Slide 7 We may want to mention its probably better to use minirouter or some other vm to control dhcp and not use the host?
- Slide 8. We support BGP now in current 2.6 build among other features
- Slide 15, BGP is already implemented
- Potentially link the following articles - /articles/ospf_cost.article, Look at the router API additional commands (about 9 routing commands)

***Module 7
- Do we want to add versioning pitfalls or troubleshooting?

GhostofGoes:
Fixes #1407

Updates the minidoc training slides per the review feedback in the issue:

- **Module 1**: use DHCP default gateway (`router dhcp <net> router <gw>`) instead of OSPF for the router sandwich
- **Module 2.5**: explain how minimega containers differ from Docker/LXC
- **Module 3**: fix module 4 link title mismatch (miniweb vs. Experiment Visualization)
- **Module 4**: reword console availability bullet
- **Module 6**: recommend minirouter over host DNSMASQ for DHCP, correct outdated "BGP planned" status (BGP is implemented), link the OSPF cost article, summarize the full router API
- **Module 7**: add a pitfall slide on miniccc/minimega version mismatches

All slide changes were verified by building and rendering `cmd/minidoc` locally.

Module 2.5 was intentionally *not* renumbered into its own module (as also suggested in the issue) to keep this change minimal and low-risk.

### DIFF
--- a/doc/content/training/module02_5.slide
+++ b/doc/content/training/module02_5.slide
@@ -49,6 +49,16 @@ A container only requires a filesystem with the various programs and files neede
 
 Consequently, containers require far fewer resources to run, and you can therefore run many more containers on a host compared to KVMs.
 
+* How minimega containers differ from Docker/LXC
+
+minimega implements its own lightweight container runtime rather than shelling out to Docker or LXC.
+
+minimega builds containers directly using Linux namespaces (mount, PID, network, UTS, IPC) and cgroups, without requiring a container engine daemon (such as dockerd) to be installed or running on the host.
+
+Because minimega manages the namespaces itself, containers are treated as first-class citizens alongside KVM VMs: they share the same `vm`config`/`vm`launch` workflow, the same networking (VLANs, taps), the same command and control (miniccc), and the same instrumentation tools (`vm`top`, packet capture, etc.).
+
+minimega containers also boot from a plain root filesystem (an `init` program and a directory tree) rather than a layered image format like a Docker image, which keeps the tooling (vmbetter) simple and avoids the overhead of an image registry or union filesystem.
+
 * Types of VM image files
 
 There are a number of image files you can use to launch a virtual machine or container. 
--- a/doc/content/training/module01.slide
+++ b/doc/content/training/module01.slide
@@ -144,9 +144,12 @@ Then, a second interface is added to the router VM, and again, the router will a
 
 * Step 3 - Configure the Router
 
-Then, we define a routing protocol of OSPF. These two lines allow the traffic of each network to cross over via the router.
+Then, we set the router as the default gateway for each network, using the `router` option of the `router`dhcp` command:
 
-Details on how OSPF works is beyond the scope of this tutorial, however we recommend the user be familiar with such networking details, if this impacts how you will be configuring your experiment.
+ router router dhcp 10.0.0.0 router 10.0.0.1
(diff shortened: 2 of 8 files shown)


doc/content/training/module02_5.slide
```diff
@@ -49,6 +49,16 @@ A container only requires a filesystem with the various programs and files neede
 
 Consequently, containers require far fewer resources to run, and you can therefore run many more containers on a host compared to KVMs.
 
+* How minimega containers differ from Docker/LXC
+
+minimega implements its own lightweight container runtime rather than shelling out to Docker or LXC.
+
+minimega builds containers directly using Linux namespaces (mount, PID, network, UTS, IPC) and cgroups, without requiring a container engine daemon (such as dockerd) to be installed or running on the host.
+
+Because minimega manages the namespaces itself, containers are treated as first-class citizens alongside KVM VMs: they share the same `vm`config`/`vm`launch` workflow, the same networking (VLANs, taps), the same command and control (miniccc), and the same instrumentation tools (`vm`top`, packet capture, etc.).
```
sdelliot, line 58: Feel free to ignore if this is some valid minidoc format:

```suggestion
Because minimega manages the namespaces itself, containers are treated as first-class citizens alongside KVM VMs: they share the same `vm`config`/`vm`launch` workflow, the same networking (VLANs, taps), the same command and control (miniccc), and the same instrumentation tools (`vm top`, packet capture, etc.).
```

sdelliot (changes requested):
This all looks good to me aside from the one possible bug. Happy to approve if that turns out to not be a bug. 

As an aside, I would love input on transitioning the documentation from the custom `.slide` and `.article` format to a reasonable standard like Markdown or RST. I plan on opening an issue for this in the future though so that the community can discuss.

doc/content/training/module02_5.slide
GhostofGoes: Good catch on the odd syntax, but this is intentional (and necessary) for the `.slide` format: minidoc's code-span parser splits on whitespace before matching backtick pairs, so ```vm top``` (with a literal space inside the backticks) never renders as code — it prints literal backticks. I verified this by rendering both forms locally.

The backtick-separated form `vm`top` (no space) is the working convention used throughout the rest of the training docs (e.g. module01.slide, module07.slide, module14.slide) to render `vm top` as a single code span. Keeping it as-is to stay consistent and correctly rendered.

sdelliot (approved):
LGTM!

Issue #1608, opened 2026-06-10
test: modernize minimega automated testing strategy


sdelliot:
## Summary

Modernize minimega testing by replacing the legacy `minitest` exact-transcript harness with a `go test`-driven test strategy that supports unit, integration, system, and cluster/distributed testing.

**Disclosure:** This issue text was drafted with AI assistance and reviewed/edited by a human collaborator before posting.

## Problem

minimega currently relies in part on a custom transcript/golden-file harness (`minitest`) that reads command files, executes them against a running daemon, writes `.got` files, and compares them against `.want` files.

While this has historically captured valuable end-to-end workflows, the current approach has several limitations:

- full-output exact matching is brittle and often breaks on formatting drift
- stale `.want` files create widespread false failures
- tests are not naturally integrated into `go test`
- the harness depends on an externally running daemon
- failures are hard to interpret and difficult to use in CI
- it is difficult to distinguish parsing/logic regressions from rendering/output changes
- distributed/cluster regressions are hard to target precisely

As a result, the current transcript suite is no longer serving well as a dependable automated signal, even though the scenario files themselves still encode useful behavioral coverage.

## Proposal

Move minimega toward a layered, Go-native testing model:

1. **Unit tests**
   - use standard Go table-driven `_test.go` tests
   - cover pure logic and parser/config behavior
   - fast, deterministic, no daemon required

2. **Integration tests**
   - run under `go test`
   - exercise command behavior through a lightweight harness
   - use structured assertions instead of exact full transcript matching

3. **System tests**
   - privileged tests for KVM, OVS, dnsmasq, etc.
   - separated from normal developer/unit runs

4. **Cluster/distributed tests**
   - explicit tests for mesh, iomeshage, scheduling, plumbing, etc.
   - run in dedicated environments or scheduled automation

5. **Transcript migration**
   - keep existing transcript files as scenario definitions / migration input
   - stop treating `.want` files as the primary long-term assertion mechanism
   - gradually port high-value transcript scenarios into structured Go tests

## Why this is valuable

This would improve:

- developer workflow (`go test ./...` becomes meaningful)
- automation readiness
- test signal quality
- debugging speed
- long-term maintainability
- confidence in distributed behavior

It would also help catch bugs at the right layer. For example, a recent diskspec-related regression only manifested in clustered environments because generic preprocessing was applied to an entire structured disk spec instead of just the path portion. Under a modern test strategy, that issue would ideally be caught by:
[...]

PR #1654, opened 2026-09-04
feat(miniccc): add client auto-update support

### LINKED ISSUE #1106: miniccc: auto-update
Rebuilding VMs to update `miniccc` is annoying. `miniccc` should do a version check when it connects to `ron` and replace itself with whatever `ron` provides.

Probably should be an option such as `cc auto-update <true,false>`. New binaries (`miniccc`, `miniccc.exe`) would be pushed from `$iomBase`.

GhostofGoes:
## Summary

- Add per-namespace `cc auto-update true|false` configuration.
- Deliver platform-specific miniccc binaries when client and RON versions differ.
- Replace and restart the client after receiving the update.
- Document the feature.

## Testing

- `go test ./internal/ron`
- Linux miniccc compilation with `GOOS=linux CGO_ENABLED=0`
- Windows miniccc compilation with `GOOS=windows CGO_ENABLED=0`

## Feedback requested

I am specifically looking for feedback on whether this remains a feature worth implementing and maintaining, including the operational and cross-platform tradeoffs.

Closes #1106

### DIFF
--- a/internal/ron/server.go
+++ b/internal/ron/server.go
@@ -56,6 +56,7 @@ type Server struct {
 	clients    map[string]*client // map of active clients, each of which have a running handler
 	vms        map[string]VM      // map of uuid -> VM
 	clientLock sync.Mutex         // lock for clients and vms
+	autoUpdate bool
 
 	path string // path for serving files
 
@@ -101,6 +102,13 @@ func NewServer(path, subpath string, plumber *miniplumber.Plumber) (*Server, err
 	log.Info("registered new ron server: %v", filepath.Join(path, subpath))
 
 	return s, nil
+}
+
+// SetAutoUpdate enables or disables miniccc binary updates on client connect.
+func (s *Server) SetAutoUpdate(enabled bool) {
+	s.clientLock.Lock()
+	defer s.clientLock.Unlock()
+	s.autoUpdate = enabled
 }
 
 func (s *Server) Destroy() {
@@ -795,6 +803,18 @@ func (s *Server) handshake(conn net.Conn, expectedUUID string) (*client, error) 
 
 	s.clients[c.UUID] = c
 
+	if s.autoUpdate && m.Client.Version != version.Revision {
+		filename := "miniccc"
+		if m.Client.OS == "windows" {
+			filename += ".exe"
+		}
+		go func() {
+			if err := s.sendUpdateFile(c, filename); err != nil {
+				log.Error("miniccc auto-update for %v: %v", c.UUID, err)
+			}
+		}()
+	}
+
 	return c, nil
 }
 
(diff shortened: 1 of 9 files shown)


nblair2:
This is could be cool, but the counterpoint is that miniccc code changes so rarely that it might not be worth using this.

internal/ron/server.go
nblair2, line 754: I would expect the auto-update logic to kick in around here?

GhostofGoes:
> This is could be cool, but the counterpoint is that miniccc code changes so rarely that it might not be worth using this.

Counter-counter point: miniccc baking sucks, mis-matched miniccc versions suck, bane of my existence

nblair2:
> > This is could be cool, but the counterpoint is that miniccc code changes so rarely that it might not be worth using this.
> 
> Counter-counter point: miniccc baking sucks, mis-matched miniccc versions suck, bane of my existence 

Is it just the log message noting mismatch that you don't like? In terms of functionality I've only seen a mismatch matter once.

Counter^3 point: maybe miniccc should be tagged with the most recent commit that changed its code and ron only reports a mismatch if there have been changes that matter to miniccc ?

sdelliot:
> > This is could be cool, but the counterpoint is that miniccc code changes so rarely that it might not be worth using this.
> 
> Counter-counter point: miniccc baking sucks, mis-matched miniccc versions suck, bane of my existence

The challenging of mismatched versions is one of the reasons why I have avoided miniccc thus far. In FIREWHEEL, we leverage just the [QEMU Guest Agent](https://sandialabs.github.io/firewheel/system/vm_resource_manager/qga.html) which seems a bit more robust. However, miniccc does have some nice features and I would like to provide that as an option as a [VM Resource Management](https://sandialabs.github.io/firewheel/system/vm_resource_manager/index.html) Driver in FIREWHEEL. We are currently adding ADB driver support (see https://github.com/sandialabs/firewheel/pull/243/) and I don't think adding miniccc would be too hard to take a similar approach.

If we do go down this path, than making sure the versions match will be a necessity. An alternative approach we could take is using the baked in QGA to drop in the correct version of miniccc, and then pivoting to use that if/when desired by users.

While technically part of FIREWHEEL, the VM Resource Manager could even be split off from the main code base as I think it could be valuable even without using some of FIREWHEEL's other high-level experiment abstractions.

nblair2:
I'm on board. I think my main qualm is trading tested setup and maintenance time (injecting new versions of miniccc into disks), which is high leverage - done once and works everywhere for experiment start time (auto upgrade), which is recurrent - done at every experiment startup.

* [ ] (can be backlog issue) update miniccc to carry a version that points to the last time it was updated, allow minimega to track that and only invoke updates if those versions diverge.
* [ ] address the comment above about the location of the log vs. these changes
* [ ] test, test, test, and re-test

GhostofGoes:
> I'm on board. I think my main qualm is trading tested setup and maintenance time (injecting new versions of miniccc into disks), which is high leverage - done once and works everywhere for experiment start time (auto upgrade), which is recurrent - done at every experiment startup.
> 
> * [ ] (can be backlog issue) update miniccc to carry a version that points to the last time it was updated, allow minimega to track that and only invoke updates if those versions diverge.
> * [ ] address the comment above about the location of the log vs. these changes
> * [ ] test, test, test, and re-test

Agree. Minimega already checks the version, so just need to upgrade that logic to be more specific to miniccc changes on either side. Changes to minimega that don't affect miniccc shouldn't necessitate an update. 

A few more items:
- This behavior should be opt-in
- On the phenix side, add a schema item to control auto updates, defaults to false. Could have an experiment wide option, per-node option, or even a phenix instance setting. Need to think about this more. 
- this is a good opportunity to drive test improvements for miniccc client binaries.

nblair2:
Yes - testing as outlined in #1608 has been on my mind. Go unit tests are good to have but complex systems like this require integration tests. I have some thoughts on how we can do this but need to look at what is possible in a GitHub runner.

PR #1667, opened 2026-09-17
fix(ron): deliver exec-once commands to clients that connect later

cmulk:
# fix(ron): deliver exec-once commands to clients that connect later

## Description ##

`cc exec-once` and `cc background-once` commands were silently dropped unless a miniccc client happened to already be connected at the moment the command was issued.

The `Once` behavior was gated by a single global `Command.Sent` flag in `internal/ron/server.go`. `sendCommands` set that flag for every command it placed into a broadcast message, *before* `route` had attempted any delivery:

```go
for k, v := range s.commands {
	if !v.Once || !v.Sent {
		m.Commands[k] = v.Copy()
		v.Sent = true          // set unconditionally, before route()
	}
}

s.route(m)                     // may deliver to nobody
```

This change replaces `Sent` with per-client delivery tracking (`Command.sentTo`) that is recorded only after a message is successfully written to the client. A `Once` command is now suppressed only for the specific clients that already received it.

The client command watermark (`client.maxCommandID`) is also now advanced from the command IDs actually sent, rather than from the maximum ID present in the batch. Previously it was assigned unconditionally, outside any success check.

Net behavior change: a `Once` command is delivered to any matching client that connects after the command was issued, while still running exactly one time per client. Non-`once` `cc exec` behavior is unchanged.

## Motivation and context ##

Reported behavior: issuing `cc exec-once` *before* the miniccc agent starts means the command is never executed. It only works if the agent is already running and connected when the command is issued. Observed on Windows guests, but the defect is entirely in the server-side `internal/ron` delivery path and is not platform-specific — Windows just makes it more visible, since miniccc running as a service typically starts later relative to when `cc` commands are scripted.

Failure sequence before this change:

1. `cc exec-once ...` sets `Once: true` (`cmd/minimega/cc_cli.go`).
2. `Server.NewCommand` stores the command and fires `go s.sendCommands("")`.
3. `sendCommands` sees `!v.Sent`, includes the command, and sets `v.Sent = true`.
4. `s.route(m)` finds `s.clients` empty. Nothing is sent. `Issued` stays `0`. Nothing rolls `Sent` back, and `Sent` was never reset anywhere in the tree.
5. The client connects; `clientHandler` calls `go s.sendCommands(c.UUID)`.
6. The gate evaluates `!v.Once || !v.Sent` → `false || false` → `false`, so the command is excluded. The client never sees it.

Because `Sent` was global rather than per client, and was set before delivery was attempted, the same flag also caused:

- **Commands dropped for disconnected clients.** `cc exec-once` targeting a set of VMs where only some were connected consumed the one-shot for all of them. The rest never ran it, even after connecting.
- **Filter mismatches consuming the one-shot.** `Sent` was set before `Client.Matches` ran, so a command whose filter excluded every connected client was still marked as sent.
- **Send failures consuming the one-shot.** A `sendMessage` error is only logged; `Sent` had already been set.
- **A startup race.** If a client was mid-handshake, `NewCommand`'s `go sendCommands("")` could run before the client was inserted into `s.clients`. Whichever of the two `sendCommands` calls ran first set `Sent`, and the other saw nothing. This is why the failure felt nondeterministic.

The only recovery was `clear cc commands` followed by re-issuing the command.

## Related Issues/PRs ##

`Once`/`Sent` was introduced in 71634c31 ("ron: add option for only executing a command once (#1508)"). 336b18cf ("feat(cc): report issued command count") made the defect observable, since an affected command shows `once=true, issued=0` in `cc commands`.

## Type of Change ##

- [x] Bugfix (`fix`)
- [ ] Feature (`feat`)
- [ ] Documentation (`docs`)
- [ ] Refactor (`refactor`)
- [ ] Chore (CI, build, dependencies, etc.) (`chore`)
- [ ] Other (please describe):

## Testing ##

Tested by running a simple experiment using startup via c2, which always (or almost always) issues the exec once commands before the VM boots. Confirmed that commands get queued and executed (and only once) when the VM boots. No changes noticed to other exec commands (they exec on every vm reboot)

Added regression tests in `internal/ron/server_test.go` covering the reported defect and the adjacent state-burning cases:

- `TestSendCommandsOnceReachesLateClient` — posts a `Once` command with **zero** clients connected, then connects a client and asserts delivery. This is the reported defect and fails before the change.
- `TestSendCommandsOnceNotResentAfterReconnect` — asserts the documented `Once` guarantee still holds: no redelivery after a reconnect resets the watermark.
- `TestSendCommandsOnceIsPerClient` — a client that was not connected at issue time still receives the command after another client already ran it.
- `TestSendCommandsFilteredCommandDoesNotBurnWatermark` — a filtered command does not report as issued, and does not make other commands unreachable.
- `TestSendCommandsOnceSurvivesSendFailure` — a failed send does not consume the one-shot; the command is delivered after the client reconnects.

The pre-existing `TestSendCommandsTracksIssuedCommands` is unmodified and still covers non-`once` replay-on-reconnect.

## Checklist ##

- [x] I have included no proprietary/sensitive information in my code or the PR.
- [x] My branch is rebased onto `master` and has one commit unless multiple commits are intentionally reviewable.
- [ ] I have tested my code (describe above). 

## Additional Notes ##

**Wire compatibility.** `Command.Sent` was an exported, gob-encoded field and is replaced by the unexported `sentTo`, which is not serialized. This is safe: `Sent` was only ever written by the server and was never read by miniccc, and gob silently ignores fields absent from the receiving struct. No miniccc update is required, and mixed-version clients are unaffected.

**Locking.** `route` now reads and writes per-command delivery bookkeeping, so its single caller (`sendCommands`) must hold `commandLock`; this is documented on both functions. Because `handleUUID` runs in parallel goroutines that share the same `*Command`, the `sentTo` reads and writes are serialized by the mutex previously named `issuedLock`, now `deliveryLock`, which already guarded `Issued`. Existing lock ordering (`commandLock` → `clientLock`, matching `ClearCommands`) is preserved.

**`Command.Copy` no longer propagates delivery state.** `Copy` feeds both the wire messages and the public `GetCommand`/`GetCommands` accessors, so it intentionally omits `sentTo`, which is internal server bookkeeping.

**Scope.** The `cc exec` (non-`once`) path is unchanged in behavior. The watermark fix does affect both paths, but only by making the watermark reflect what was actually delivered.


GhostofGoes:
Thanks for this fix, I've run into this with the past with some scorch components and never took the time to look into it. A persistent source of miniccc "weirdness" that will now be resolved.

GhostofGoes (approved):

